PR #23, opened 2022-06-09
Remove cosign pkg, take a dep on it.

vaikas:

#### Summary
Remove the pkg/ directories that have to do with verification, aka, cosign bits. Take a dependency
on it rather than even try to maintain two parallel implementations. Then as time passes, some of
these dependencies will make their way to sigstore/sigstore and some won't, but at least it's very
clear now where we take deps to cosign.

#### Ticket Link

Fixes

#### Release Note

```release-note

```


imjasonh (approved):
This looks great! It looks like there aren't any remaining deps on `cmd/cosign` at least, which is great. We should put in a test to ensure no new ones sneak in. 👍

pkg/webhook/validation.go
```diff
@@ -27,11 +27,11 @@ import (
 	"knative.dev/pkg/apis"
 	"knative.dev/pkg/logging"
 
-	"github.com/sigstore/policy-controller/cmd/cosign/cli/fulcio/fulcioroots"
+	"github.com/sigstore/cosign/cmd/cosign/cli/fulcio/fulcioroots"
```
imjasonh, line 30: This seems to be the only cmd/cosign dependency, and it's moving to sigstore/sigstore in https://github.com/sigstore/sigstore/pull/435 🎉

vaikas: Yeah, can't wait!

vaikas:
Once we get rid of the fulcioroots, I'll add that test, ok?

cpanato (approved):
amazing

PR #25, opened 2022-06-09
Add spdxjson and cyclonedx as supported predicate types

jdolitsky:
Please see the following PRs in cosign where these new types are added:

- https://github.com/sigstore/cosign/pull/1974
- https://github.com/sigstore/cosign/pull/1977

### DIFF
--- a/pkg/apis/policy/v1alpha1/clusterimagepolicy_validation.go
+++ b/pkg/apis/policy/v1alpha1/clusterimagepolicy_validation.go
@@ -165,7 +165,7 @@ func (a *Attestation) Validate(ctx context.Context) *apis.FieldError {
 	}
 	if a.PredicateType == "" {
 		errs = errs.Also(apis.ErrMissingField("predicateType"))
-	} else if a.PredicateType != "custom" && a.PredicateType != "slsaprovenance" && a.PredicateType != "spdx" && a.PredicateType != "link" && a.PredicateType != "vuln" {
+	} else if a.PredicateType != "custom" && a.PredicateType != "slsaprovenance" && a.PredicateType != "spdx" && a.PredicateType != "spdxjson" && a.PredicateType != "cyclonedx" && a.PredicateType != "link" && a.PredicateType != "vuln" {
 		// TODO(vaikas): The above should be using something like:
 		// if _, ok := options.PredicateTypeMap[a.PrecicateType]; !ok {
 		// But it causes an import loop. That refactor can be part of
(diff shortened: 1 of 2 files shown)


pkg/apis/policy/v1alpha1/clusterimagepolicy_validation.go
```diff
@@ -165,7 +165,7 @@ func (a *Attestation) Validate(ctx context.Context) *apis.FieldError {
 	}
 	if a.PredicateType == "" {
 		errs = errs.Also(apis.ErrMissingField("predicateType"))
-	} else if a.PredicateType != "custom" && a.PredicateType != "slsaprovenance" && a.PredicateType != "spdx" && a.PredicateType != "link" && a.PredicateType != "vuln" {
+	} else if a.PredicateType != "custom" && a.PredicateType != "slsaprovenance" && a.PredicateType != "spdx" && a.PredicateType != "spdxjson" && a.PredicateType != "cyclonedx" && a.PredicateType != "link" && a.PredicateType != "vuln" {
```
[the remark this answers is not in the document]
jdolitsky: good call. it hath been done

pkg/apis/policy/v1beta1/clusterimagepolicy_validation.go
```diff
@@ -165,7 +169,7 @@ func (a *Attestation) Validate(ctx context.Context) *apis.FieldError {
 	}
 	if a.PredicateType == "" {
 		errs = errs.Also(apis.ErrMissingField("predicateType"))
-	} else if a.PredicateType != "custom" && a.PredicateType != "slsaprovenance" && a.PredicateType != "spdx" && a.PredicateType != "link" && a.PredicateType != "vuln" {
+	} else if !validPredicateTypes.Has(a.PredicateType) {
 		// TODO(vaikas): The above should be using something like:
```
imjasonh, line 173: lol I didn't even see this comment before. Is the import loop still a problem? It'd be nice not to have to maintain this list in two places.

imjasonh: Sounds like that's https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/options/predicate.go#L40

The good news is, I don't think that would cause a dependency cycle.

The bad news is, it's a new dependency on cmd/cosign 😭 

I think we could probably just maintain two sets for now, and maybe move both into sigstore/sigstore (or into cosign/pkg/something). Doesn't need to block this PR though I think.

imjasonh (approved):

pkg/apis/policy/v1beta1/clusterimagepolicy_validation.go
vaikas: Yeah, +1 for just having a local copy and move it somewhere else. But, yes, agree on not being part of this PR. We should remove this TODO from here though, since its' addressed by the TODO above. But that can be done in a follow on as well, I'll do it 👍

vaikas (approved):

PR #29, opened 2022-06-11
Fix #28. Treat multiple authorities as OR instead of AND.

### LINKED ISSUE #28: Multiple authorities are treated as AND instead of OR
**Description**

When specifying both `key` and `keyless` section in the ClusterImagePolicies, the webhook enforces both of them instead of one of them.

**Version**

vaikas:

#### Summary
Fix #28 by treating multiple authorities as OR instead of AND. This was what it was supposed to be as per the spec all along.

#### Ticket Link

Fixes

#### Release Note

```release-note
* Fix bug #28 where multiple authorities were treated as AND instead of OR as specified.
```

### DIFF
--- a/pkg/webhook/validator.go
+++ b/pkg/webhook/validator.go
@@ -262,7 +262,7 @@ func validatePolicies(ctx context.Context, namespace string, ref name.Reference,
 			result := retChannelType{name: cipName}
 
 			result.policyResult, result.errors = ValidatePolicy(ctx, namespace, ref, cip, remoteOpts...)
-			if len(result.errors) == 0 {
+			if len(result.policyResult.AuthorityMatches) > 0 {
 				// Ok, at least one Authority  on the policy passed. If there's a CIP level
 				// policy, apply it against the results of the successful Authorities
 				// outputs.
@@ -366,7 +366,8 @@ func ValidatePolicy(ctx context.Context, namespace string, ref name.Reference, c
 		}()
 	}
 	// If none of the Authorities for a given policy pass the checks, gather
-	// the errors here. If one passes, do not return the errors.
+	// the errors here. Even if there are errors, return the matched
+	// authoritypolicies.
 	authorityErrors := []error{}
 	// We collect all the successfully satisfied Authorities into this and
 	// return it.
@@ -392,9 +393,9 @@ func ValidatePolicy(ctx context.Context, namespace string, ref name.Reference, c
 			}
 		}
 	}
-	if len(authorityErrors) > 0 {
-		return nil, authorityErrors
-	}
+	// Even if there are errors, return the policies, since as per the
+	// spec, we just need one signature to pass checks. If more than
+	// one are required, that is enforced at the policy level.
 	return policyResult, authorityErrors
 }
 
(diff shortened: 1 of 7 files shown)


vaikas:
latest run, setup ko failed with:
```
+ case latest-release in
++ curl -s https://api.github.com/repos/google/ko/releases/latest
++ jq -r .tag_name
+ tag=null
Installing ko @ null for Linux
+ os=Linux
+ [[ Linux == \m\a\c\O\S ]]
+ [[ ! -z null ]]
+ echo 'Installing ko @ null for Linux'
+ sudo tar xzf - -C /usr/local/bin ko
+ curl -fsL https://github.com/google/ko/releases/download/null/ko_ull_Linux_x86_[64](https://github.com/sigstore/policy-controller/runs/6841937796?check_suite_focus=true#step:5:66).tar.gz
gzip: stdin: unexpected end of file
tar: Child returned status 1
tar: Error is not recoverable: exiting now
Error: Process completed with exit code 2.
```

I'll retry that if others work.

test/testdata/policy-controller/e2e/cip-keyless-with-attestations.yaml
```diff
@@ -33,8 +33,3 @@ spec:
         data: |
           predicateType: "cosign.sigstore.dev/attestation/v1"
           predicate: Data: "foobar e2e test"
```
hectorj2f, line 35: Why are you removing this?

hectorj2f: Is it intentional?

pkg/webhook/validator.go
```diff
@@ -262,21 +262,23 @@ func validatePolicies(ctx context.Context, namespace string, ref name.Reference,
 			result := retChannelType{name: cipName}
 
 			result.policyResult, result.errors = ValidatePolicy(ctx, namespace, ref, cip, remoteOpts...)
-			if len(result.errors) == 0 {
```
hectorj2f, line 265: Suggestion: The line `cip := cip` doesn’t make sense to me.

hectorj2f (approved):

test/testdata/policy-controller/e2e/cip-keyless-with-attestations.yaml
vaikas: Yeah, so with the move to OR, it's not really meaningful in our tests. So, in the tests now, we create first the keyless.yaml (or key.yaml) which check for signature, then the key[less]-with-attestations.yaml adds the AND by creating a new CIP so modified the tests to check for that as well. 
And cip-key-and-keyless.yaml tests the OR capability.
Otherwise, we'd have to bake the 'AND' (requires both signature and attestation) into the CIP policy, and I'd rather leave that to other PRs where we create better examples of CUE policies. I created #31 to track that issue.
@hectorj2f does that make sense?

dlorenc (approved):

cpanato (approved):

vaikas:
I don't know why I didn't see this in the PR review. Reason for those are
to declare locals inside the for loop since we're launching a go func. Does
that make sense?

test/testdata/policy-controller/e2e/cip-keyless-with-attestations.yaml
hectorj2f: Yes, I understood it now. We could create more examples. Thanks

Issue #59, opened 2022-06-29
Certain malformed policies are rejected at enforcement time instead of validation


mattmoor:
**Description**

If you have a bad policy, e.g. `glob: $FOO*`, our validation isn't smart enough to catch this during the CIP validation, however, this logic at enforcement time catches it and rejects all the images as a result:
https://github.com/sigstore/policy-controller/blob/2ef684f8f0e5ea41d3483a0a84ea906c33d4ecbf/pkg/apis/config/glob.go#L50-L67

This is the failure mode I see (snipped for readability):
```
... denied the request: validation failed: invalid glob "$FOO*": spec.containers[0].image
index.docker.io/library/ubuntu@sha256:b6b83d3c331794420340093eb706a6f152d9c1fa51b262d9bf34594887c2c7ac
```

I'll take a crack at this now, but wanted to open an issue to track things.

PR #60, opened 2022-06-29
Validate globs at admission time.

mattmoor:
This move the logic for compiling (incl. validating) and matching globs out of `pkg/apis/config` into `pkg/apis/glob` (to avoid a dep cycle) and adds logic to attempt `glob.Compile` when validating ClusterImagePolicy CRDs to avoid invalid CIPs rejecting requests due to invalid policies *at enforcement time*.

Fixes: https://github.com/sigstore/policy-controller/issues/59

#### Release Note
```release-note
ClusterImagePolicy validation is now more consistent about checking glob validity when applied (vs. failing during enforcement)
```

### DIFF
--- a/pkg/apis/policy/v1beta1/clusterimagepolicy_validation.go
+++ b/pkg/apis/policy/v1beta1/clusterimagepolicy_validation.go
@@ -58,14 +58,10 @@ func (spec *ClusterImagePolicySpec) Validate(ctx context.Context) (errors *apis.
 }
 
 func (image *ImagePattern) Validate(ctx context.Context) *apis.FieldError {
-	var errs *apis.FieldError
 	if image.Glob == "" {
-		errs = errs.Also(apis.ErrMissingField("glob"))
-	} else {
-		errs = errs.Also(ValidateGlob(image.Glob).ViaField("glob"))
+		return apis.ErrMissingField("glob")
 	}
-
-	return errs
+	return ValidateGlob(image.Glob).ViaField("glob")
 }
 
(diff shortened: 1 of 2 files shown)


mattmoor:
cc @vaikas @hectorj2f

pkg/apis/policy/v1beta1/clusterimagepolicy_validation.go
```diff
@@ -60,10 +61,10 @@ func (image *ImagePattern) Validate(ctx context.Context) *apis.FieldError {
 	var errs *apis.FieldError
 	if image.Glob == "" {
 		errs = errs.Also(apis.ErrMissingField("glob"))
+	} else {
```
hectorj2f, line 64: nit:  It looks like we could have a return in the if statement `if image.Glob == "" { return errs.Also(apis.ErrMissingField("glob"))` so we don't need the following else statement.

hectorj2f (dismissed):
lgtm, just a nit

pkg/apis/policy/v1beta1/clusterimagepolicy_validation.go
mattmoor: Yeah, I went with this way so that we could collect more errors below the bit I made an `else`, but since this is just for globs, IDK that there'll be much else.

The new `glob.Compile` was just piling on when complaining about empty-string, which I thought was unnecessarily confusing.  I'll switch these to returns instead.

mattmoor: Should be RFAL 🙏

PR #67, opened 2022-07-05
Add policy-tester CLI for testing ClusterImagePolicies

jdolitsky:
Adding a simple tool for validating ClusterImagePolicy resources against various images on the command line

### DIFF
--- a/cmd/tester/main.go
+++ b/cmd/tester/main.go
@@ -74,11 +74,12 @@ func main() {
 
 	// TODO(jdolitsky): This should use v1beta1 once there exists a
 	// webhookcip.ConvertClusterImagePolicyV1beta1ToWebhook() method
-	var tmp v1alpha1.ClusterImagePolicy
-	if err := yaml.Unmarshal(cipRaw, &tmp); err != nil {
+	var v1alpha1cip v1alpha1.ClusterImagePolicy
+	if err := yaml.Unmarshal(cipRaw, &v1alpha1cip); err != nil {
 		log.Fatal(err)
 	}
-	cip := webhookcip.ConvertClusterImagePolicyV1alpha1ToWebhook(&tmp)
+	v1alpha1cip.SetDefaults(ctx)
+	cip := webhookcip.ConvertClusterImagePolicyV1alpha1ToWebhook(&v1alpha1cip)
 	ref, err := name.ParseReference(*image)
 	if err != nil {
 		log.Fatal(err)
--- a/cmd/tester/main.go
+++ b/cmd/tester/main.go
@@ -1,0 +1,101 @@
+//
+// Copyright 2022 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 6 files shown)


cmd/tester/main.go
```diff
@@ -0,0 +1,101 @@
...
+	}())
+)
+
+type output struct {
+	Errors []string              `json:"errors"`
+	Result *webhook.PolicyResult `json:"result"`
+}
+
+func main() {
+	args := os.Args[1:]
```
vaikas, line 61: Could we make these proper flags instead of positional arguments?

cmd/tester/main.go
```diff
@@ -0,0 +1,101 @@
...
+		os.Exit(1)
+	}
+	cipFilePath := args[0]
+	image := args[1]
+	cipRaw, err := ioutil.ReadFile(cipFilePath)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// TODO(jdolitsky): should this use v1beta1?
```
vaikas, line 73: yep :)

test/testdata/policy-controller/tester/cip-public-keyless.yaml
```diff
@@ -0,0 +1,24 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: policy.sigstore.dev/v1alpha1
+kind: ClusterImagePolicy
+metadata:
+  name: demo
+spec:
+  images:
+    - glob: "**"
```
vaikas, line 21: I wonder if we could maybe point this at say cosign release instead?

vaikas (commented):
Sweet! Thanks for doing this.

cmd/tester/main.go
```diff
@@ -0,0 +1,101 @@
...
+	var tmp v1alpha1.ClusterImagePolicy
+	if err := yaml.Unmarshal(cipRaw, &tmp); err != nil {
+		log.Fatal(err)
+	}
+	cip := webhookcip.ConvertClusterImagePolicyV1alpha1ToWebhook(&tmp)
```
vaikas, line 78: I wonder if it would make sense to call the defaults on the policy before this. Just thinking that way it's simpler to write the test policies so that you don't have to fill all the fields. Just a thought :)
https://github.com/sigstore/policy-controller/blob/main/pkg/apis/policy/v1beta1/clusterimagepolicy_defaults.go#L22

cmd/tester/main.go
jdolitsky: So.. it doesnt seem like there is a way to convert the v1beta1 resource to webhook type. I've updated the TODO to reflect that

jdolitsky:
One thing I noticed here is that the actual image doesn't necessarily need to match the glob pattern in order to test it... is that something we should check in this tool too? (can come later)

vaikas (approved):
Thanks!!!

hectorj2f (approved):

PR #69, opened 2022-07-06
Add static field to authority that allows for block/allow behaviour.

vaikas:

#### Summary
Add `static` as a sibling to `key` and `keyless` in authority block. `static` has an `action` field that currently is either 'pass' or 'fail'. This is not a bool for now, since I reckon we could implement 'warn' for example later on, allow it but warn. This is a k8s webhook way to signal warnings to users.

https://github.com/sigstore/cosign/issues/1653

#### Release Note

* Adds a new field to authority called static that can be used to explicitly block or allow certain images.

#### Documentation

https://github.com/sigstore/sigstore-website/issues/188


hectorj2f (approved):

MageshSrinivasulu:
@hectorj2f Have we implemented the action "warn"? If so can you please add the documentation here https://github.com/sigstore/policy-controller in the readme.md it hard to navigate and read through all the issues

vaikas:
@MageshSrinivasulu we haven't implemented it yet, but do want to as soon as we get time :)

PR #74, opened 2022-07-12
Validate CIP before using it.

vaikas:

#### Summary
Validate CIP after defaulting and before using it to catch mistakes in authoring a CIP.


hectorj2f (approved):
lgtm

mattmoor (approved):
Wondering if we should we `SetDefaults` as well? 🤔

vaikas:
@mattmoor it does call SetDefaults right above it?

PR #76, opened 2022-07-13
Set reinvocationPolicy to 'IfNeeded' for the tag resolver webhook

vpnachev:
#### Summary

Set reinvocationPolicy to 'IfNeeded' for the tag resolver webhook. This is needed when other mutating webhooks alter pod spec, e.g. inject a sidecar container, and result into images using tags.

ref: https://sigstore.slack.com/archives/C01PZKDL4DP/p1657691580474789


hectorj2f (approved):
lgtm

PR #80, opened 2022-07-13
Remove replace directives by upgrading etcd to v3.6.0-alpha.0.

wlynch:
#### Summary

The replace directive makes it harder for others to import the packages, since it can
break minimum version selection, causing issues when other packages
import this module.

We're already importing this package else where in sigstore
https://github.com/sigstore/cosign/blob/a7c439a29e96d01afee18918d0466ceb685dd747/go.mod#L267.

Signed-off-by: Billy Lynch <billy@chainguard.dev>

#### Release Note

NONE


hectorj2f (commented):
I’d suggest to avoid using alpha versions. We won’t be able to use the policy controller while it uses alphas.

wlynch:
> We won’t be able to use the policy controller while it uses alphas.

Can you explain this a bit more? if this is a hard blocker for policy-controller this might be a time bomb since the next release of cosign will cause this upgrade for etcd. 😬 

For the time being, reworked this PR by downgrading the otel deps to the replaced version.

hectorj2f:
@wlynch It is not a blocker but security advocates for avoiding using any dependency pointing to an alpha version. That is the problem.

wlynch:
Soon is not looking promising :( https://github.com/etcd-io/etcd/issues/13538#issuecomment-1095112286

I'll go ahead and close this PR - anyone who wants to import this module will just need to copy the replace statements for the time being 😞 

I'll follow up on https://github.com/sigstore/cosign/pull/2070 about what we want to do about the cosign's usage of the alpha etcd package.

Thanks!

PR #88, opened 2022-07-14
Fix #38 , do not block status updates.

### LINKED ISSUE #38: Old versions of deployment with untrusted image cannot be teared down
**Description**

I have an already running deployment which image is not signed. Cosign has been enabled for the namespace of the deployment and I would like to update the deployment with another image that is signed can be successfully verified by the policy-controller. 
So, the change was accepted for the `Deployment` resource, new `ReplicaSet` was created and then the first `Pod` is also created. At this point, kube-controller-manager tries to scale down the old replica set, however changing the old `ReplicaSet` is disallowed by cosign as it is still using the untrusted image. 

Steps to reproduce (assuming cosign is installed and running). 
1. `kubectl create namespace test`
2. 
```bash
cat <<EOF | kubectl -n test create -f -
apiVersion: apps/v1
kind: Deployment
metadata:
  name: foo-bar
spec:
  replicas: 2
  selector:
    matchLabels:
      k8s-app: foo-bar
  strategy:
    rollingUpdate:
      maxSurge: 1
      maxUnavailable: 0
    type: RollingUpdate
  template:
    metadata:
      labels:
        k8s-app: foo-bar
    spec:
      containers:
      - command: 
        - "sh"
        - "-c"
        - "sleep 3600"
        image: alpine:3.16
        imagePullPolicy: IfNotPresent
        name: sleep
EOF
```
3. `kubectl label namespace test policy.sigstore.dev/include=true`
4. `kubectl -n test patch deployments.apps foo-bar -p '{"spec":{"template":{"spec":{"containers":[{"name":"sleep","image":"<my-signed-image>"}]}}}}'`

After several seconds, the state of deployment will stuck with new replicaset with 1 replica and old replica set that cannot be scaled down
```bash
kubectl -n test get deployment,replicaset,pod
NAME                      READY   UP-TO-DATE   AVAILABLE   AGE
deployment.apps/foo-bar   3/2     1            3           8m

NAME                                 DESIRED   CURRENT   READY   AGE
replicaset.apps/foo-bar-7f7559865f   2         2         2       8m
replicaset.apps/foo-bar-bc97b45d     1         1         1       5m

NAME                           READY   STATUS    RESTARTS   AGE
pod/foo-bar-7f7559865f-d99v8   1/1     Running   0          8m
pod/foo-bar-7f7559865f-qgfjh   1/1     Running   0          8m
pod/foo-bar-bc97b45d-ndr4d     1/1     Running   0          5m
```

The error in the policy-controller is
```json
{
  "level": "error",
  "ts": "2022-06-14T19:23:57.097Z",
  "logger": "policy-controller",
  "caller": "validation/validation_admit.go:170",
  "msg": "Failed the resource specific validation",
  "commit": "a4cb262",
  "knative.dev/kind": "apps/v1, Kind=ReplicaSet",
  "knative.dev/namespace": "test",
  "knative.dev/name": "foo-bar-7f7559865f",
  "knative.dev/operation": "UPDATE",
  "knative.dev/resource": "apps/v1, Resource=replicasets",
  "knative.dev/subresource": "",
  "knative.dev/userinfo": "{system:serviceaccount:kube-system:deployment-controller a1132716-21af-4e79-9d5c-ff0440b2953d [system:serviceaccounts system:serviceaccounts:kube-system system:authenticated] map[]}",
  "error": "no matching signatures:\n: spec.template.spec.containers[0].image\n<my-signed-image-digest>",
  "stacktrace": "knative.dev/pkg/webhook/resourcesemantics/validation.validate\n\tknative.dev/pkg@v0.0.0-20220325200448-1f7514acd0c2/webhook/resourcesemantics/validation/validation_admit.go:170\nknative.dev/pkg/webhook/resourcesemantics/validation.(*reconciler).Admit\n\tknative.dev/p
kg@v0.0.0-20220325200448-1f7514acd0c2/webhook/resourcesemantics/validation/validation_admit.go:80\nknative.dev/pkg/webhook.admissionHandler.func1\n\tknative.dev/pkg@v0.0.0-20220325200448-1f7514acd0c2/webhook/admission.go:117\nnet/http.HandlerFunc.ServeHTTP\n\tnet/http/server.go:2047\
nnet/http.(*ServeMux).ServeHTTP\n\tnet/http/server.go:2425\nknative.dev/pkg/webhook.(*Webhook).ServeHTTP\n\tknative.dev/pkg@v0.0.0-20220325200448-1f7514acd0c2/webhook/webhook.go:262\nknative.dev/pkg/network/handlers.(*Drainer).ServeHTTP\n\tknative.dev/pkg@v0.0.0-20220325200448-1f7514
acd0c2/network/handlers/drain.go:110\nnet/http.serverHandler.ServeHTTP\n\tnet/http/server.go:2879\nnet/http.(*conn).serve\n\tnet/http/server.go:1930"
}
```

I am not sure how this can be avoided when the policy-controller is validating not only pods but also the higher level resources like replicasets and deployments. Also, it looks like the KCM does not use the `/scale` subresource which is not subject of admission control by the cosign and would allow tearing down the old replicaset. Here is sample patch requests not using and using the `/scale` subresource:
```bash
$ kubectl -n test patch replicaset foo-bar-7f7559865f  -p '{"spec":{"replicas": 1}}'
Error from server (BadRequest): admission webhook "policy.sigstore.dev" denied the request: validation failed: no matching signatures:
: spec.template.spec.containers[0].image
index.docker.io/library/alpine@sha256:4edbd2beb5f78b1014028f4fbb99f3237d9561100b6881aabbf5acce2c4f9454

$ kubectl -n test patch replicaset foo-bar-7f7559865f  -p '{"spec":{"replicas": 1}}' --subresource=scale
scale.autoscaling/foo-bar-7f7559865f patched
```

**Version**

```json
{
  "gitVersion": "v1.9.0",
  "gitCommit": "a4cb262dc3d45a283a6a7513bb767a38a2d3f448",
  "gitTreeState": "clean",
  "buildDate": "2022-06-03T13:47:07Z",
  "goVersion": "go1.17.11",
  "compiler": "gc",
  "platform": "linux/amd64"
}
```

vaikas:
#### Summary
If a scalable resource (deployment, rs, etc.) is not-compliant, allow for it to scale down. Fixes #38 
While playing around with this, noticed also that we are in a non-compliant case _also_ blocking status updates, which is pretty bad, so tackling that here as well.

#### Release Note

* Fixes #38 which was preventing non-compliant scalable resources (like deployment) from being torn down / scaled down
* Also allows status updates even if non-compliant.


pkg/webhook/validator.go
```diff
@@ -574,9 +620,39 @@ func ValidatePolicyAttestationsForAuthority(ctx context.Context, ref name.Refere
...
+
+	if apis.IsInUpdate(ctx) {
+		newReplicaCount := ps.Spec.Replicas
+		original := apis.GetBaseline(ctx).(*policyduckv1beta1.PodScalable)
+		if newReplicaCount != nil && original != nil && original.Spec.Replicas != nil {
+			if *newReplicaCount < *original.Spec.Replicas {
+				logging.FromContext(ctx).Debugf("Skipping validations due to scale down request %s/%s", &ps.ObjectMeta.Name, &ps.ObjectMeta.Namespace)
+				return
+			}
+		}
```
mattmoor, line 638: Maybe ad `IsScalingDown(ctx)` on `PodScalable` that you can call here?

pkg/webhook/validator.go
```diff
@@ -574,9 +620,39 @@ func ValidatePolicyAttestationsForAuthority(ctx context.Context, ref name.Refere
...
+		ServiceAccountName: ps.Spec.Template.Spec.ServiceAccountName,
+		ImagePullSecrets:   imagePullSecrets,
+	}
+	v.resolvePodSpec(ctx, &ps.Spec.Template.Spec, opt)
+}
+
 // ResolvePodSpecable implements duckv1.PodSpecValidator
 func (v *Validator) ResolvePodSpecable(ctx context.Context, wp *duckv1.WithPod) {
-	if wp.DeletionTimestamp != nil {
+	if apis.IsInDelete(ctx) {
```
mattmoor, line 655: You should check that `IsInDelete` checks the deletion timestamp, since we don't want to mess with things being finalized either.

pkg/apis/duck/v1beta1/podscalable_types.go
```diff
@@ -0,0 +1,113 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"knative.dev/pkg/apis"
+	"knative.dev/pkg/apis/duck"
+	"knative.dev/pkg/ptr"
+)
+
+// +genduck
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// PodScalable is a duck type that the are PodSpecable but also can scale.
```
hectorj2f, line 30: Question: should we add this new type to knative duck types ?

hectorj2f (approved):

pkg/apis/duck/v1beta1/podscalable_types.go
vaikas: Since it's fairly "easy" to add new duck types, Knative folks have been little hesitant to keep adding them to pkg repo. I could see about pushing them upstream later on though.

hectorj2f: Alright! 👍🏻

hectorj2f (approved):

PR #97, opened 2022-07-22
fix path in cloudbuild

cpanato:
#### Summary
- fix path in cloudbuild

#### Release Note

`n/a`

#### Documentation

`n/a`


hectorj2f (approved):
good catch!

PR #100, opened 2022-07-22
chore: add golang documentation to the api types

hectorj2f:
Signed-off-by: hectorj2f <hectorf@vmware.com>

#### Summary

I added some missing documentation about the purpose of each field in the ClusterImagePolicy definition.

### DIFF
--- a/config/300-clusterimagepolicy.yaml
+++ b/config/300-clusterimagepolicy.yaml
@@ -48,11 +48,13 @@ spec:
               type: object
               properties:
                 authorities:
+                  description: Authorities defines the rules for discovering and validating signatures.
                   type: array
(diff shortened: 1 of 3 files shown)


vaikas:
Can you also add the field descriptions to v1alpha1? Then you need to run ./hack/update-codegen.sh so it will update the actual CRD in config/ directory.

JAORMX (approved):

vaikas:
Looks like there's something wrong with the updatecode gen because it's not updating any of the comments in the v1beta1, only on v1alpha1. But, that's something that we can fix after this. Thanks!

vaikas (approved):

PR #108, opened 2022-07-27
Begin an examples/ directory

jdolitsky:


vaikas:
This is a great start. My original thinking was that these would live somewhere that we could also run tests against them. Here's couple of them:
https://github.com/sigstore/policy-controller/tree/main/test/testdata/policies

I think that having examples is more discoverable name for sure, but I'd also like to add tests to make sure that we don't break examples for folks. WDYT?

vaikas:
Sorry (point about them being discoverable ;) ), here:

https://github.com/sigstore/policy-controller/tree/main/test/testdata/policy-controller/e2e

jdolitsky:
Since the example is using keyless fulcio sig, I've set it to run only on push to main

For other examples (using custom keypair), we could run the in PRs

vaikas (approved):

hectorj2f (approved):

PR #115, opened 2022-07-29
Fix keyful handling in tester.

vaikas:

#### Summary
@jdolitsky found in #108 that the tester was not handling the Keyful validation correctly. This was caused by us not 'round-tripping' the webhook cip, which in the Policy-Controller path is marshalled by the reconciler and unmarshalled before handing to the validator path. One thing the unmarshaller does is convert the inlined data blocks into crypto keys which is what the policy-controller validator uses.

So, this PR adds that round tripping before using the webhook cip.

#### Release Note
* Fix bug in tester which was not handling keyful signatures / attestations correctly.


mattmoor (approved):
We may want to add a public method we can call on the Go version to do this without the roundtrip, but I think the comment makes this silliness clear enough for now.

hectorj2f (approved):

Issue #109, opened 2022-07-28
Use a wrapped error type to distinguish between transient errors and policy violations


mattmoor:
**Description**

Today `ValidatePolicy` returns `(*PolicyResult, []error)`, but the errors are a somewhat opaque mixture of:
- Real policy failures
- Transient and non-policy failures

I think it would be extremely useful to be able to distinguish between actual policy violations, and other (potentially transient, or misconfiguration) issues that aren't necessarily policy violations, but user or dependent service errors (e.g. GCR serves a 500).

Recent versions of Go have nice error type wrapping constructs that we can use, and which support wrapping the errors as they propagate up, so I'd like to propose something along the lines of:

```go
type PolicyViolation struct { ... }

// PolicyViolation implements error
var _ error = (*PolicyViolation)(nil)
```

... which would allow callers of `ValidatePolicy` to write something like:
```go

pr, errs := ValidatePolicy(...)
for _, err := range errs {
   pv := &PolicyViolation{}
   if errors.As(err, &pv) {
      // We have a PolicyViolation, react accordingly!
   } else {
      // Some other error, let's handle it differently!
   }
}
```

We could take this a step further, and use this to surface `PolicyWarning` as a separate ~level of error.  We could then potentially use this to allow users to configure the `policy-controller` behavior on non-policy errors similar to the `{Validating,Mutating}WebhookConfiguration` offers the `failurePolicy` for configuring the behavior of Kubernetes webhooks when the webhook isn't responding as expected (e.g. it is down).

cc @hectorj2f @vaikas

mattmoor:
Hmm, this actually needs changes to cosign as well, since this calls into that for signature and policy validation.

PR #116, opened 2022-07-30
Wrap errors, encode signature Subject/Issuer.

mattmoor:
As I was poking around the code base preparing for the upstream wrapped error change to land, I refactored a bunch of code paths (in part preparing to integrate that change).

There are two main changes:
1. Pull in the wrapped error changes,
2. Drop the use of `PLACEHOLDER` and encode the actual `Subject`/`Issuer` from the `oci.Signature`.

There are a bunch of minor / supporting changes:
 - collapse several redundant code paths,
 - pre-allocate several arrays and maps to avoid resizing,
 - make some of the error messages more consistent across sig/att and key/keyless,
 - use the wrapped errors a handful of places.

Fixes: https://github.com/sigstore/policy-controller/issues/109


mattmoor:
cc @vaikas

mattmoor:
FML, hopefully this lands: https://github.com/google/certificate-transparency-go/pull/962

mattmoor:
I've added a `replace` for @jdolitsky change here to see if CI finds any other problems, but we need that change to land, or to do some work in `sigstore/cosign` to unblock the flow of changes 😞

pkg/webhook/validation.go
```diff
@@ -43,14 +42,7 @@ func valid(ctx context.Context, ref name.Reference, rekorClient *client.Rekor, k
...
 		}
-		sps, err := validSignaturesWithFulcio(ctx, ref, fulcioRoots, nil /* rekor */, nil /* no identities */, opts...)
-		if err != nil {
-			return nil, err
-		}
-		if len(sps) > 0 {
-			return sps, nil
-		}
-		return nil, errors.New("no valid signatures were found")
+		return validSignaturesWithFulcio(ctx, ref, fulcioRoots, nil /* rekor */, nil /* no identities */, opts...)
```
mattmoor, line 45: It turns out that when `sps` is empty, this always returns an error, so the flow we had was pointless.

pkg/webhook/validator.go
```diff
@@ -460,12 +463,22 @@ func ValidatePolicy(ctx context.Context, namespace string, ref name.Reference, c
...
-	ret := []PolicySignature{}
+	ret := make([]PolicySignature, 0, len(sigs))
 	for _, ociSig := range sigs {
 		logging.FromContext(ctx).Debugf("Converting signature %+v", ociSig)
-		ret = append(ret, PolicySignature{Subject: "PLACEHOLDER", Issuer: "PLACEHOLDER"})
+
+		if cert, err := ociSig.Cert(); err == nil && cert != nil {
+			ce := cosign.CertExtensions{
+				Cert: cert,
+			}
```
mattmoor, line 473: This is extracted directly from how `cosign` outputs this information during verification today.

pkg/webhook/validator.go
```diff
@@ -494,13 +507,11 @@ func ValidatePolicySignaturesForAuthority(ctx context.Context, ref name.Referenc
...
 		sps, err := valid(ctx, ref, rekorClient, authority.Key.PublicKeys, remoteOpts...)
 		if err != nil {
-			return nil, fmt.Errorf("failed to validate public keys with authority %s for %s: %w", name, ref.Name(), err)
-		} else if len(sps) > 0 {
-			logging.FromContext(ctx).Debugf("validated signature for %s with authority %s got %d signatures", ref.Name(), authority.Name, len(sps))
-			return ociSignatureToPolicySignature(ctx, sps), nil
+			return nil, fmt.Errorf("key validation failed for authority %s for %s: %w", name, ref.Name(), err)
 		}
-		logging.FromContext(ctx).Errorf("no validSignatures found with authority %s for %s", name, ref.Name())
-		return nil, fmt.Errorf("no valid signatures found with authority %s for %s", name, ref.Name())
```
mattmoor, line 503: As I said above, this is unreachable today.

pkg/webhook/validator.go
```diff
@@ -511,18 +522,16 @@ func ValidatePolicySignaturesForAuthority(ctx context.Context, ref name.Referenc
...
 			if err != nil {
 				logging.FromContext(ctx).Errorf("failed validSignatures for authority %s with fulcio for %s: %v", name, ref.Name(), err)
-				return nil, fmt.Errorf("validate signatures with fulcio: %w", err)
-			} else if len(sps) > 0 {
-				logging.FromContext(ctx).Debugf("validated signature for %s, got %d signatures", ref.Name(), len(sps))
-				return ociSignatureToPolicySignature(ctx, sps), nil
+				return nil, fmt.Errorf("keyless validation failed for authority %s for %s: %w", name, ref.Name(), err)
 			}
-			logging.FromContext(ctx).Errorf("no validSignatures found for %s", ref.Name())
-			return nil, fmt.Errorf("no valid signatures found with authority %s for  %s", name, ref.Name())
```
mattmoor, line 520: same

pkg/webhook/validator.go
```diff
@@ -567,23 +576,25 @@ func ValidatePolicyAttestationsForAuthority(ctx context.Context, ref name.Refere
 			va, err := validAttestationsWithFulcio(ctx, ref, fulcioroot, rekorClient, authority.Keyless.Identities, remoteOpts...)
 			if err != nil {
 				logging.FromContext(ctx).Errorf("failed validAttestationsWithFulcio for authority %s with fulcio for %s: %v", name, ref.Name(), err)
-				return nil, fmt.Errorf("validate signatures with fulcio: %w", err)
```
mattmoor, line 570: Similar to the API, I thought it prudent to avoid the term "fulcio" in error messages, especially since the API keyword is `keyless`.  I also tried to align this with the more detailed errors for the `key` path above.

pkg/webhook/validator.go
```diff
@@ -567,23 +576,25 @@ func ValidatePolicyAttestationsForAuthority(ctx context.Context, ref name.Refere
...
 		}
 	}
+
 	// If we didn't get any verified attestations either from the Key or Keyless
 	// path, then error out
 	if len(verifiedAttestations) == 0 {
-		logging.FromContext(ctx).Errorf("no valid attestations found with authority %s for %s", name, ref.Name())
-		return nil, fmt.Errorf("no valid attestations found with authority %s for %s", name, ref.Name())
+		logging.FromContext(ctx).Errorf("no valid attestations found for authority %s for %s", name, ref.Name())
+		return nil, fmt.Errorf("%w for authority %s for %s", cosign.ErrNoMatchingAttestations, name, ref.Name())
```
mattmoor, line 589: Pretty sure this too is unreachable, but since we have the loop and all I opted for wrapping the same error here instead.

pkg/webhook/validator.go
```diff
@@ -382,55 +384,56 @@ func ValidatePolicy(ctx context.Context, namespace string, ref name.Reference, c
 			switch {
 			case authority.Static != nil:
 				if authority.Static.Action == "fail" {
-					result.err = errors.New("disallowed by static policy")
+					result.err = cosign.NewVerificationError("disallowed by static policy")
 					results <- result
 					return
 				}
-				result.signatures = []PolicySignature{{Subject: "allowed by static policy", Issuer: "allowed by static policy"}}
```
mattmoor, line 389: This was necessary because of the `default` case below, which errors when both `result.signatures` and `result.attestations` were empty.  I've changed that to allow this case, so that we don't have to look for these non-`const` sentinel strings.

pkg/webhook/validator.go
```diff
@@ -382,55 +384,56 @@ func ValidatePolicy(ctx context.Context, namespace string, ref name.Reference, c
...
+
 			case len(result.attestations) > 0:
 				policyResult.AuthorityMatches[result.name] = AuthorityMatch{Attestations: result.attestations}
+
 			default:
-				authorityErrors = append(authorityErrors, fmt.Errorf("failed to process authority: %s", result.name))
+				// This happens when we encounter a policy with:
+				//   static:
+				//     action: "pass"
+				policyResult.AuthorityMatches[result.name] = AuthorityMatch{}
```
mattmoor, line 436: This is the `default` case I was referring to above.  We hit this when there wasn't an error, but there also weren't any validates `signatures` or `attestations`, which corresponds to the `static: { action: pass }` path.

mattmoor (commented):
Left some comments inline to explain things, which probably wouldn't make sense as committed code comments.

pkg/webhook/validator.go
```diff
@@ -511,18 +522,16 @@ func ValidatePolicySignaturesForAuthority(ctx context.Context, ref name.Referenc
...
-			return nil, fmt.Errorf("no valid signatures found with authority %s for  %s", name, ref.Name())
+			logging.FromContext(ctx).Debugf("validated signature for %s, got %d signatures", ref.Name(), len(sps))
+			return ociSignatureToPolicySignature(ctx, sps), nil
 		}
 	}
+
 	// This should never happen because authority has to have been
 	// validated to be either having a Key or Keyless
-	return nil, fmt.Errorf("authority has neither key, keyless, or static specified")
+	return nil, errors.New("authority has neither key, keyless, or static specified")
```
vaikas, line 534: while changing this, mind fixing the comment above that I failed to update :)

pkg/webhook/validator.go
```diff
@@ -552,11 +561,11 @@ func ValidatePolicyAttestationsForAuthority(ctx context.Context, ref name.Refere
 			va, err := validAttestations(ctx, ref, verifier, rekorClient, remoteOpts...)
 			if err != nil {
 				logging.FromContext(ctx).Errorf("error validating attestations: %v", err)
-				return nil, fmt.Errorf("validating attestations: %w", err)
+				return nil, fmt.Errorf("key validation failed for authority %s for %s: %w", name, ref.Name(), err)
```
vaikas, line 564: looks like maybe copy&paste from signature validation, can you add something here about the attestation.

pkg/webhook/validator.go
```diff
@@ -567,23 +576,25 @@ func ValidatePolicyAttestationsForAuthority(ctx context.Context, ref name.Refere
 			va, err := validAttestationsWithFulcio(ctx, ref, fulcioroot, rekorClient, authority.Keyless.Identities, remoteOpts...)
 			if err != nil {
 				logging.FromContext(ctx).Errorf("failed validAttestationsWithFulcio for authority %s with fulcio for %s: %v", name, ref.Name(), err)
-				return nil, fmt.Errorf("validate signatures with fulcio: %w", err)
+				return nil, fmt.Errorf("keyless validation failed for authority %s for %s: %w", name, ref.Name(), err)
```
vaikas, line 579: ditto.

pkg/webhook/validator.go
```diff
@@ -382,55 +384,56 @@ func ValidatePolicy(ctx context.Context, namespace string, ref name.Reference, c
...
+	for range cip.Authorities {
 		select {
 		case <-ctx.Done():
-			authorityErrors = append(authorityErrors, fmt.Errorf("context was canceled before validation completed"))
+			authorityErrors = append(authorityErrors, fmt.Errorf("%w before validation completed", ctx.Err()))
+
 		case result, ok := <-results:
 			if !ok {
-				authorityErrors = append(authorityErrors, fmt.Errorf("results channel failed to produce a result"))
+				authorityErrors = append(authorityErrors, errors.New("results channel closed before all results were sent"))
```
vaikas, line 419: just for my own edification, is that really the only reason why that would return !ok?

pkg/webhook/validator.go
vaikas: I see what you're doing here but having a default below that is fail open just makes me a bit nervous (not in this PR), but in general.
What do you think about changing this structure here and add a field for Static pass instead:
https://github.com/sigstore/policy-controller/blob/main/pkg/webhook/validator_result.go#L40

Say, AuthorityMatch.StaticPass

Then we could tag this as having satisfied that and check for that before the default clause and leave default be fail?

Just thinking again for future and safety?

pkg/webhook/validation.go
vaikas: superduper.

pkg/webhook/validator.go
```diff
@@ -26,9 +26,11 @@ import (
 	"github.com/google/go-containerregistry/pkg/authn/k8schain"
 	"github.com/google/go-containerregistry/pkg/name"
 	"github.com/google/go-containerregistry/pkg/v1/remote"
+	"github.com/sigstore/cosign/pkg/cosign"
 	"github.com/sigstore/cosign/pkg/oci"
 	ociremote "github.com/sigstore/cosign/pkg/oci/remote"
 	"github.com/sigstore/cosign/pkg/policy"
+	csigs "github.com/sigstore/cosign/pkg/signature"
```
vaikas, line 33: We've been trying (with the sigstore refactoring work that @imjasonh and others have been doing) to reduce dependencies to cosign and where ever possible to use sigstore/sigstore, so I'm not sure we want to add this one here. I can also see us maybe having additional error types that we define so maybe we could just have our Errors that we return just like you did in sigstore/cosign and use them here instead.

vaikas (commented):
Thanks for doing this, great find on the sps never being 0 👍 Also, thanks for filling in the placeholder!

pkg/webhook/validator.go
mattmoor: I get that, but here it is a little chicken/egg.  I only added the error stuff to cosign because that's where the logic still lives (and that should be trivial to move to `sigstore/sigstore` when the signing/verification moves).

I'd have to look at what it would take to move the extension stuff over, but it'd be great to have a library in `sigstore/sigstore` that all of `cosign`, `fulcio`, and `policy-controller` can lean on 🤞

pkg/webhook/validator.go
mattmoor: As long as we're not lying about it having signatures, I'm happy 😁

pkg/webhook/validator.go
mattmoor: Yup:
image

pkg/webhook/validator.go
mattmoor: Yup, done!

pkg/webhook/validator.go
mattmoor: I prefixed the signature paths with `signature` and the attestation paths with `attestation` to distinguish.  I guess I made them too similar 🤣

pkg/webhook/validator.go
vaikas: Yeah, that's why think having a new field that's not lying is good. I probably should've just added it the first time around :)

pkg/webhook/validator.go
vaikas: Thanks lazygithub :)

pkg/webhook/validator.go
vaikas: Ack, that makes sense so that when the signing moves the errors would move as well. Hindsight, maybe the error stuff should've been put into sigstore/sigstore to begin with?

mattmoor: Possibly, but I opted for the faster/lazier version of treating it as part of the verification chunk that'll move later 😏

pkg/webhook/validation.go
hectorj2f: 👍🏻

pkg/webhook/validator.go
hectorj2f: yes, it makes sense.

vaikas:
LGTM, thanks much!

hectorj2f:
lgtm, thanks for the PR.

mattmoor:
NP, I'll rebase once Josh's PR merges 🤞

mattmoor:
Just rebased on @jdolitsky change 🙏

jdolitsky (approved):

vaikas (approved):

Issue #107, opened 2022-07-27
Make subject/issuer required


vaikas:
**Description**

TL;DR not specifying issuer/subject is bad, we should not allow folks to craft CIPs that do not specify those. More details here:
https://docs.google.com/document/d/1o8_bXIygufgiohJGlmBzqF4_BnXCTfgh4ILgJFJxYRs/edit?resourcekey=0-YEar3v67uoT31kj83dCVvA#heading=h.oiw6nn1ucgaq

hectorj2f:
@vaikas I started working on this, do we want to make required the subject and issuer on any keyless configuration ?

hectorj2f:
https://github.com/sigstore/policy-controller/pull/125 I opened a draft PR,
do we want to require the issuer and subject whenever a CaCert is not defined ?

vaikas:
That's a great question and I was wondering about that as well. From the original design document, it was described as this, but I honestly don't think this has even been implemented or documented or tested?
```
# For air-gapped scenarios, we support identifying the
      # relevant keys local to the cluster, either via inline
      # public-key data, or a secretRef (in the cosign-system
      # namespace).  Each of these share a shape with the
      # "key" field above.
      ca-cert:
	   name: # arbitrary identifier
        data: # inline public key data
```
... time passes ... Yeah, so it's not documented, nor is it implemented. Maybe we should ask some of the folks from the task force for what the purpose / usage was and how does it play with keyless. Maybe @mattmoor remembers?
I kind of feel that if it's not implemented, we might want to actually reject that field now, because it's a poor user experience, or we should implement it and add tests for it.

mattmoor:
re: Issuer/Subject

Given that the API is now beta and adding required fields is breaking, can we add a "Warning" for at least a release or two when they aren't specified before making them required?

re: unimplemented fields

I'd reject anything we don't support outright vs. silently doing anything.  We can always relax things later as we add support.  For this (since it doesn't *really* work) I'd be fine with going straight to errors.

vaikas:
This was fixed in:
https://github.com/sigstore/policy-controller/pull/125

vaikas:
I'm leaving this out for the second question about the cacert behaviour however, since I don't think that rejection was implemented.

hectorj2f:
Closing this issue! It has been recently implemented in main.

PR #125, opened 2022-08-03
chore: require setting the identity - issuer and subject

hectorj2f:
Signed-off-by: hectorj2f <hectorf@vmware.com>

#### Summary

related to #107 

I am just checking what we want to enforce with this Draft pull request, some e2e tests need to be adapted to this new validation.

### DIFF
--- a/pkg/apis/policy/v1beta1/clusterimagepolicy_validation.go
+++ b/pkg/apis/policy/v1beta1/clusterimagepolicy_validation.go
@@ -172,15 +172,9 @@ func (keyless *KeylessRef) Validate(ctx context.Context) *apis.FieldError {
 	if keyless.CACert != nil {
 		errs = errs.Also(keyless.DeepCopy().CACert.Validate(ctx).ViaField("ca-cert"))
 	}
-	// Throw an error in case there are not a defined identity and a ca-cert for a keyless configuration
-	if len(keyless.Identities) == 0 && keyless.CACert == nil {
-		errs = errs.Also(apis.ErrMissingField("identities"))
-	} else {
-		for i, identity := range keyless.Identities {
-			errs = errs.Also(identity.Validate(ctx).ViaFieldIndex("identities", i))
-		}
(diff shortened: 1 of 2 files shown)


hectorj2f:
I am working on adding warnings instead of errors following https://github.com/knative/pkg/pull/2498

pkg/apis/policy/v1beta1/clusterimagepolicy_validation.go
```diff
@@ -226,7 +228,7 @@ func (p *Policy) Validate(ctx context.Context) *apis.FieldError {
 func (identity *Identity) Validate(ctx context.Context) *apis.FieldError {
 	var errs *apis.FieldError
 	if identity.Issuer == "" && identity.Subject == "" && identity.IssuerRegExp == "" && identity.SubjectRegExp == "" {
-		errs = errs.Also(apis.ErrMissingField("issuer", "subject", "issuerRegExp", "subjectRegExp"))
+		return apis.ErrMissingField("issuer", "subject", "issuerRegExp", "subjectRegExp")
```
hectorj2f, line 231: i doesn't make sense to evaluate the fields if all are empty.

vaikas (approved):
Awesomesauce! Thanks for doing this! We should maybe add a section for these in our e2e tests as well. I added some code that you can create a CIP and add an expected error that gets validated. Here's where it's checked:
https://github.com/sigstore/policy-controller/blob/main/test/e2e_test_policy_crd.sh#L38

A sample entry for an error here:
https://github.com/sigstore/policy-controller/blob/main/test/testdata/policy-controller/invalid/keylessref-with-empty-authorities.yaml#L15

So, maybe we can add test/testdata/policy-controller/warn directory and add some CIPs there that produce warnings, but instead of ERROR: maybe have a section like WARN: and make sure the warning comes through.

WDYT? We could do that obvs in a followup.

hectorj2f:
@vaikas it sounds like a plan to me 👍🏻

Issue #130, opened 2022-08-04
Cue policy in test file for two attestations does not work.


vaikas:
**Description**
The following is a CIP policy json that's being evaluated:
```
{
  "authorityMatches": {
    "keyatt": {
      "signatures": null,
      "attestations": {
        "custom-match-predicate": [
          {
            "subject": "",
            "issuer": ""
          }
        ]
      }
    },
    "keylesssignature": {
      "signatures": [
        {
          "subject": "https://kubernetes.io/namespaces/default/serviceaccounts/default",
          "issuer": "https://kubernetes.default.svc/"
        }
      ],
      "attestations": null
    },
    "keysignature": {
      "signatures": [
        {
          "subject": "",
          "issuer": ""
        }
      ],
      "attestations": null
    }
  }
}
```

It is incorrectly passing with this policy file here:
https://github.com/sigstore/policy-controller/blob/main/test/testdata/policy-controller/e2e/cip-requires-two-signatures-and-two-attestations.yaml
And in particular here:
https://github.com/sigstore/policy-controller/blob/main/test/testdata/policy-controller/e2e/cip-requires-two-signatures-and-two-attestations.yaml#L95

**Version**

hectorj2f:
I believe this has been fixed!

vaikas:
Awesome, is this test passing if you uncomment this?
https://github.com/sigstore/policy-controller/blob/main/test/e2e_test_cluster_image_policy_with_attestations.sh#L229

vaikas:
This may help here:
https://tip.cuelang.org/play/?id=--Ep9vOrwU1#cue@export@cue

PR #134, opened 2022-08-04
use omit-empty to make results more readable and concise.

vaikas:

#### Summary
While debugging #130 we realized that there's some cruft there, so let's get rid of it to make things more readable.


hectorj2f (approved):

mattmoor:
thank you! 🙏

PR #135, opened 2022-08-05
Include the predicate type and payload for attestations.

mattmoor:
This change factors a new `PolicyAttestation` type so that we can encode additional information for attestations beyond just signature fields.

This change adds the predicate type (which would otherwise require parsing the CIP to know), and the payload (included for callers of ValidatePolicy, so they don't have to refetch the `oci.Signature`'s `Payload()`).

#### Release Note

The policy results for attestations now includes the predicate type (available to CIP policies) and the attestation payload (not available to CIP policies since we've already run the policy over them).  To achieve this, there is a potentially breaking change to the signature of `ValidatePolicy` to wrap the attestation's `PolicySignature` in a new `PolicyAttestation` type, however, to minimize disruption `PolicyAttestation` inlines `PolicySignature`.

#### Documentation

cc @hectorj2f @vaikas

### DIFF
--- a/pkg/webhook/validator.go
+++ b/pkg/webhook/validator.go
@@ -498,6 +498,10 @@ func ociSignatureToPolicySignature(ctx context.Context, sigs []oci.Signature) []
 	return ret
 }
 
+// attestation is used to accumulate the signature along with extracted and
+// validated metadata during validation to construct a list of
+// PolicyAttestations upon completion without needing to refetch any of the
+// parts.
 type attestation struct {
 	oci.Signature
 
@@ -505,7 +509,7 @@ type attestation struct {
 	Payload       []byte
 }
 
-func attestationToPolicyAttestation(ctx context.Context, atts []attestation) []PolicyAttestation {
+func attestationToPolicyAttestations(ctx context.Context, atts []attestation) []PolicyAttestation {
 	ret := make([]PolicyAttestation, 0, len(atts))
 	for _, att := range atts {
 		logging.FromContext(ctx).Debugf("Converting attestation %+v", att)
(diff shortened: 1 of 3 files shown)


pkg/webhook/validator.go
```diff
@@ -490,8 +490,53 @@ func ociSignatureToPolicySignature(ctx context.Context, sigs []oci.Signature) []
...
-			ret = append(ret, PolicySignature{})
+			ret = append(ret, PolicySignature{
+				// TODO(mattmoor): Is there anything we should encode for key-based?
+			})
+		}
+	}
+	return ret
+}
+
+type attestation struct {
```
vaikas, line 501: It might be good to document why this intermediate is necessary for shuffling things back and forth, it's not obvious why it's necessary.

pkg/webhook/validator.go
```diff
@@ -490,8 +490,53 @@ func ociSignatureToPolicySignature(ctx context.Context, sigs []oci.Signature) []
...
+	oci.Signature
+
+	PredicateType string
+	Payload       []byte
+}
+
+func attestationToPolicyAttestation(ctx context.Context, atts []attestation) []PolicyAttestation {
```
vaikas, line 508: nit: since it takes array, I'd think attestationToPolicyAttestations would be more appropriate.

pkg/webhook/validator.go
mattmoor: added a comment

pkg/webhook/validator.go
mattmoor: done

vaikas:
Oh, and I fully acknowledge I screwed up the original ociSignatureToPolicySignature func name, that's why it was on my mind :)

vaikas (approved):

hectorj2f (approved):
Lgtm

PR #157, opened 2022-08-17
Add cosign initialize as an init container to initialize TUF.

vaikas:

#### Documentation
* Policy Controller now uses TUF root that gets initialized in an init container.

### DIFF
--- a/.github/workflows/kind-cluster-image-policy.yaml
+++ b/.github/workflows/kind-cluster-image-policy.yaml
@@ -43,7 +43,7 @@ jobs:
 
     env:
       KO_DOCKER_REPO: "registry.local:5000/policy-controller"
-      SCAFFOLDING_RELEASE_VERSION: "v0.4.3"
+      SCAFFOLDING_RELEASE_VERSION: "v0.4.4"
       GO111MODULE: on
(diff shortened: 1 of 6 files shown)


vaikas:
Hm, so cosign initialize is failing, but looks right to me:
```
cosign initialize --mirror http://tuf.tuf-system.172.18.255.1.sslip.io/ --root ./root.json
```

```
+ cosign initialize --mirror http://tuf.tuf-system.172.18.255.1.sslip.io/ --root ./root.json
Error: creating cached local store: stat root.json/tuf.db: not a directory
main.go:62: error during command execution: creating cached local store: stat root.json/tuf.db: not a directory
Error: Process completed with exit code 1.
```

vaikas:
Ahaha, the TUF_ROOT env variable is used by sigstore/sigstore TUF to figure out where the root lives. Argh, that means I have to do some mungery until I can fix that in scaffolding.

cpanato (approved):

vaikas:
Actually, after looking more deeply at how cosign does it, I do wonder if I should actually use the code from sigstore/sigstore/tuf directly and not bake in the init container. Please hold up on merging. I'll create a different PR to see if it makes more sense.

PR #159, opened 2022-08-17
Use TUF instead of env variables.

vaikas:
#### Summary
Add two new flags that you can use to specify how to initialize TUF root.
--tuf-mirror
--tuf-root

We then do equivalent of cosign initialize with those.

#### Release Note
* Add two new flags that allow you to bring in a custom TUF root: --tuf-mirror and --tuf-root.
* DEPRECATED: Enviromental variables should be considered deprecated as the whole sigstore is moving everything to TUF.

### DIFF
--- a/.github/workflows/kind-cluster-image-policy.yaml
+++ b/.github/workflows/kind-cluster-image-policy.yaml
@@ -43,7 +43,7 @@ jobs:
 
     env:
       KO_DOCKER_REPO: "registry.local:5000/policy-controller"
-      SCAFFOLDING_RELEASE_VERSION: "v0.4.3"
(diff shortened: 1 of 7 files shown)


vaikas:
If all works here, I'd like to use this instead of #157 so that we don't need to bring in the additional initContainer.

cpanato (approved):

hectorj2f (dismissed):
lgtm

vaikas:
Thanks so much @cpanato for making kustomize work, super neat!

hectorj2f (approved):

vaikas:
Documentation changes here:
https://github.com/sigstore/sigstore-website/pull/214

PR #162, opened 2022-08-18
add missing action to install ko

cpanato:
#### Summary
- add missing action to install ko

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -23,6 +23,7 @@ jobs:
 
       - uses: sigstore/cosign-installer@b3413d484cc23cf8778c3d2aa361568d4eb54679 # v2
       - uses: anchore/sbom-action/download-syft@bb716408e75840bbb01e839347cd213767269d4a # v0.11.0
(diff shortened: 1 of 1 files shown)


hectorj2f (approved):

.github/workflows/release.yaml
```diff
@@ -23,6 +23,7 @@ jobs:
 
       - uses: sigstore/cosign-installer@b3413d484cc23cf8778c3d2aa361568d4eb54679 # v2
       - uses: anchore/sbom-action/download-syft@bb716408e75840bbb01e839347cd213767269d4a # v0.11.0
+      - uses: imjasonh/setup-ko@78eea08f10db87a7a23a666a4a6fe2734f2eeb8d # v0.4
```
hectorj2f, line 26: question: why do we need ko now ?

cpanato: before we were using cloudbuild and the builder image had that, now we change to run on github actions so we need to add that

vaikas (approved):

.github/workflows/release.yaml
hectorj2f: thanks for the explanation.

PR #188, opened 2022-08-24
Add warn tests for creating CIP with missing identities.

vaikas:

#### Summary
Add tests for making sure that when we create CIPs that warnings are produced and they match what we expect.


cpanato (approved):
Cool

PR #195, opened 2022-08-28
update CIP to fix tests when running in the push to main and in keyless mode

cpanato:
#### Summary
- update CIP to fix tests when running in the push to main and in keyless mode

```
main.go:93: Using the following cip:
kind: ClusterImagePolicy
metadata:
  creationTimestamp: null
  name: keyless-attestation-sbom-spdxjson
spec:
  authorities:
  - attestations:
    - name: must-have-spdxjson
      policy:
        data: |
          predicateType: "https://spdx.dev/Document"
        type: cue
      predicateType: spdxjson
    keyless:
      url: https://fulcio.sigstore.dev/
    name: keyless
  images:
  - glob: '**'
  mode: enforce
main.go:97: CIP is invalid: missing field(s): spec.authorities[0].keyless.identities
Error: Process completed with exit code 1.
```

see https://github.com/sigstore/policy-controller/runs/8058669693?check_suite_focus=true

### DIFF
--- a/examples/policies/keyless-attestation-sbom-spdxjson.yaml
+++ b/examples/policies/keyless-attestation-sbom-spdxjson.yaml
@@ -32,6 +32,9 @@ spec:
   - name: keyless
     keyless:
       url: "https://fulcio.sigstore.dev"
+      identities:
+        - issuer: https://token.actions.githubusercontent.com
+          subject: "https://github.com/sigstore/policy-controller/.github/workflows/policy-tester-examples.yml@refs/heads/main"
     attestations:
     - name: must-have-spdxjson
       predicateType: spdxjson


examples/policies/keyless-attestation-sbom-spdxjson.yaml
```diff
@@ -32,6 +32,9 @@ spec:
   - name: keyless
     keyless:
       url: "https://fulcio.sigstore.dev"
+      identities:
```
vaikas, line 35: Hm, so this is a fine thing to included in the examples, thank you! However, the fact that tests are not passing seems to indicate that there's a bug somewhere since missing identities should really be `warning` and not an error, for example here:
https://github.com/sigstore/policy-controller/pull/188

Let me go look.

vaikas (commented):
Hm. So this fix is fine, but I think there might be another bug somewhere, since these should be warnings and not errors.

examples/policies/keyless-attestation-sbom-spdxjson.yaml
vaikas: Yeah, there was a bug :)

cpanato: ok! cool

vaikas:
@cpanato Sorry for being unclear, I don't think we should close this. Since these are examples, I believe we should have our examples reflect the best practices, and while missing identities today are warnings, they will be required in not too long of a future, so I will re-open and let's get this in :) But, I can't reopen since you deleted the brnach, would you mind reopening? :)

cpanato:
ahh
got it will add it back @vaikas

PR #196, opened 2022-08-28
Do not exit on warnings on CIP.

vaikas:

#### Summary
As seen on #195 the tester would exit on `warning` level validation errors. So, split out the warnings and print them
separately and do not exit on them.


cpanato (approved):
cool

PR #199, opened 2022-08-29
Fix webhook looking for credentials in the wrong namespace

### LINKED ISSUE #143: Webhook looking for credentials in the wrong namespace
**Description**

I found a bug where the webhook is looking for credentials in the wrong namespace when a pod is deployed to a non-default namespace. The console error is `Error from server (BadRequest): admission webhook "policy.sigstore.dev" denied the request: validation failed: invalid value: nginx must be an image digest: spec.containers[0].image`, which is misleading to users. Looking at the logs I found the real issue:
```
{"level":"warn","ts":"2022-08-10T14:25:50.824Z","logger":"policy-controller","caller":"webhook/validator.go:733","msg":"Unable to build k8schain: secrets \"dummy-registry\" not found","commit":"6e9b0b8","knative.dev/kind":"/v1, Kind=Pod","knative.dev/namespace":"my-secure-ns","knative.dev/name":"nginx","knative.dev/operation":"CREATE","knative.dev/resource":"/v1, Resource=pods","knative.dev/subresource":"","knative.dev/userinfo":"xxxx"}
```
**Steps to reproduce**
1. Install policy controller either from the Helm charts or from the development repo.
2. Create the following policy:
```
kind: ClusterImagePolicy
metadata:
  name: image-policy
spec:
  images:
  - glob: "gcr.io/projectsigstore/cosign*"
  authorities:
  - name: official-cosign-key
    key:
      data: |
        -----BEGIN PUBLIC KEY-----
        MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEhyQCx0E9wQWSFI9ULGwy3BuRklnt
        IqozONbbdbqz11hlRJy9c7SG+hdcFl9jE9uE/dwtuwU2MqU9T/cN0YkWww==
        -----END PUBLIC KEY-----
```
3. Create a namespace enabled for verification and a docker secret:
```
apiVersion: v1
kind: Namespace
metadata:
  labels:
    policy.sigstore.dev/include: "true"
  name: my-secure-ns
---
apiVersion: v1
kind: Secret
metadata:
  name: dummy-registry
  namespace: my-secure-ns
type: kubernetes.io/dockerconfigjson
data:
  .dockerconfigjson: eyJhdXRocyI6eyJnaGNyLmlvIjp7InVzZXJuYW1lIjoiam9obmRvZSIsInBhc3N3b3JkIjoiMTIzNDU2NzgiLCJhdXRoIjoiYW05b2JtUnZaVG94TWpNME5UWTNPQT09In19fQ==
``` 
4. Patch the default service account in our new namespace:
```
kubectl patch serviceaccount default -n my-secure-ns -p '{"imagePullSecrets": [{"name": "dummy-registry"}]}'
```
5. Try to run a pod using an imperative command:
```
kubectl run nginx --image=nginx -n my-secure-ns
```
6. You should see the error now. Deploying a pod that matches the policy results in the same error:
```
kubectl run cosign -n my-secure-ns  \
--image=gcr.io/projectsigstore/cosign:v1.2.1 \
--restart=Never \
--command -- sleep 900
```

**Notes**
Deploying the pod using a yaml or a deployment does not show the error, it succeeds:
```
apiVersion: v1
kind: Pod
metadata:
  name: cosign
  namespace: my-secure-ns
spec:
  containers:
  - command:
    - sleep
    - "900"
    image: gcr.io/projectsigstore/cosign:v1.2.1
    name: cosign
```

**Version**

The bug has been reproduced in v0.2.1 in GKE and OpenShift.

elfotografo007:
#### Summary

Extract the namespace from the `AdmissionRequest` if the resource doesn't have the namespace field set correctly. Fixes #143.
#### Release Note

* Fixes #143. Webhook was incorrectly rejecting pods deployed in a non-default namespace when `imagePullSecrets` were involved.

### DIFF
--- a/pkg/webhook/validator_test.go
+++ b/pkg/webhook/validator_test.go
@@ -23,6 +23,7 @@ import (
 	"crypto/elliptic"
 	"crypto/x509"
 	"encoding/base64"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"net/http"
@@ -44,10 +45,12 @@ import (
 	policyduckv1beta1 "github.com/sigstore/policy-controller/pkg/apis/duck/v1beta1"
 	"github.com/sigstore/policy-controller/pkg/apis/policy/v1alpha1"
 	webhookcip "github.com/sigstore/policy-controller/pkg/webhook/clusterimagepolicy"
+	admissionv1 "k8s.io/api/admission/v1"
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"knative.dev/pkg/apis"
 	duckv1 "knative.dev/pkg/apis/duck/v1"
 	fakekube "knative.dev/pkg/client/injection/kube/client/fake"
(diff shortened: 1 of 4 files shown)


pkg/webhook/validator_test.go
```diff
@@ -1844,3 +1847,720 @@ func TestValidatePoliciesCancelled(t *testing.T) {
...
+	}
+	kc.CoreV1().ServiceAccounts("my-secure-ns").Patch(ctx, "default", types.MergePatchType, patch, metav1.PatchOptions{})
+
+	v := NewValidator(ctx, secretName)
+
+	cvs := cosignVerifySignatures
+	defer func() {
+		cosignVerifySignatures = cvs
+	}()
+	// Let's just say that everything is verified.
```
vaikas, line 1954: I'm curious if this actually tests that we use the imagepullsecrets from the correct namespace because we use these stubbed out ones. Or if it was modified it's hard to see because this looks like cut&paste from the above tests. Would it maybe be enough to override one of these 'pass' or 'fail' and check in them that the checkopts have indeed the correct secret referenced in them.

elfotografo007: I am not sure if we can test it there because most of the errors I saw were raised when trying to call `k8schain.New` while passing `k8schain.Options` without the proper namespace.

vaikas (approved):
Discussed offline and validated that these new tests will fail if the fix is not in.
Thanks @elfotografo007 !!

PR #219, opened 2022-09-06
remove double quotes, looks like it is passing as a single string to cosign and not as an array

cpanato:
#### Summary
- `ko` now output more digests (before 0.11.2 release the output was just one) and that was passing as a unique big string to cosign failing the job


vaikas (approved):

cpanato:
hope this is the last one 😓

PR #221, opened 2022-09-08
Allow fetching CIPs from URLs.

vaikas:

#### Summary
Sometimes it's handy to not fetch the CIP from the file, but for example by hosting the CIP for example on a website.
So, treat the -policy so that if it starts with https:// or http:// it will fetch it.

Example invocation:
```
./tester -policy="https://raw.githubusercontent.com/sigstore/policy-controller/main/examples/policies/signed-by-github-actions.yaml" -image=gcr.io/projectsigstore/policy-webhook@sha256:d1e7af59381793687db4673277005276eb73a06cf555503138dd18eaa1ca47d6
```

Which fails with the expected, issuer does not match because tester uses main, but the released Subject is:
```
https://github.com/sigstore/policy-controller/.github/workflows/release.yaml@refs/tags/v0.3.0
```

### DIFF
--- a/cmd/tester/main.go
+++ b/cmd/tester/main.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"log"
+	"net/http"
 	"os"
 	"strings"
 
@@ -63,7 +64,7 @@ type output struct {
 }
 
 func main() {
-	cipFilePath := flag.String("policy", "", "path to ClusterImagePolicy")
+	cipFilePath := flag.String("policy", "", "path to ClusterImagePolicy or URL to fetch from (http/https)")
 	image := flag.String("image", "", "image to compare against policy")
 	flag.Parse()
 	if *cipFilePath == "" || *image == "" {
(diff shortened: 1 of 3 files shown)


cpanato (approved):
cool

hectorj2f (dismissed):
lgtm, we should consider using sget :) in the future

cmd/tester/main.go
```diff
@@ -63,17 +64,32 @@ type output struct {
...
-		log.Fatal(err)
+	var cipRaw []byte
+	var err error
+	if strings.HasPrefix(*cipFilePath, "https://") || strings.HasPrefix(*cipFilePath, "http://") {
+		log.Printf("Fetching CIP from: %s", *cipFilePath)
+		resp, err := http.Get(*cipFilePath)
+		if err != nil {
+			log.Fatal(err)
+		}
+		defer resp.Body.Close()
```
hectorj2f, line 83: You could move this code to a function so the defer statements get executed, and then log a Fatal error here.

hectorj2f (approved):

PR #223, opened 2022-09-08
Remove secret name flag

### LINKED ISSUE #222: Remove the -secret-name flag
**Description**

We need to remove the `-secret-name` flag from the installation of the webhook. This would require opening PRs in the policy-controller, helm-charts and sigstore-website repositories. Likewise we agreed on deleting the current validation, so it only relies any defined ClusterImagePolicies to admit/deny container images. 
In other words, we are deleting this deprecated behaviour that was explained here https://github.com/sigstore/policy-controller#deprecated-policy-controller-validation-behavior.

hectorj2f:
#### Summary

This PR aims to remove the -secret-name flag and get rid off the deprecated behaviour (see more details here https://github.com/sigstore/policy-controller#deprecated-policy-controller-validation-behavior) that validated containers images against the fulcio root and the public key specified in the flag.

closes #222

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,8 +31,6 @@ are `OR`.
 
 See the [Configuring Image Pattern](#configuring-image-patterns) for more information.
 
-If no policy is matched against the image digest, the [deprecated policy-controller validation behavior](#deprecated-policy-controller-validation-behavior) will occur.
-
 An example of an allowed admission would be:
 1. If the image matched against `policy1` and `policy3`
 1. A valid signature or attestation was obtained for `policy1` with at least one of the `policy1` authorities
@@ -45,12 +43,7 @@ An example of a denied admission would be:
 1. No valid signature or attestation was obtained for `policy2` with at least one of the `policy2` authorities
 1. The image is not admitted
 
-An example of no policy matched:
-1. If the image does not match against any policy
-1. Fallback to [deprecated policy-controller validation behavior](#deprecated-policy-controller-validation-behavior)
-1. Validation will be attempted against the fulcio root.
-  1. If a valid signature or attestation is obtained, image is admitted
-  1. If no valid signature or attestation is obtained, image is denied
+**If an image does not match a policy, it will be admitted**
 
 ### Configuring policy-controller ClusterImagePolicy
 
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ An example of a denied admission would be:
 An example of no policy matched:
 1. If the image does not match against any policy
 1. Fallback to [deprecated policy-controller validation behavior](#deprecated-policy-controller-validation-behavior)
-1. Validation will be attempted against the secret defined under `cosign.secretKeyRef.name` during helm installation.
+1. Validation will be attempted against the fulcio root.
   1. If a valid signature or attestation is obtained, image is admitted
   1. If no valid signature or attestation is obtained, image is denied
 
@@ -376,29 +376,6 @@ spec:
         }
       }
 ```
-
-### Deprecated policy-controller Validation Behavior
-
-**Note:** This behavior is being deprecated in favor of using `ClusterImagePolicy` resources.
-
-During the admission validation, if no `ClusterImagePolicy` is matched, the deprecated behavior will occur.
-Image digests will be validated against the public key secret defined by `cosign.secretKeyRef.name` during installation.
-If the public key secret is not configured, the admission validation verifies against the fulcio root.
-
-When installing `policy-controller` through helm, `cosign.secretKeyRef.name` can be specified.
-```bash
-helm install policy-controller -n cosign-system sigstore/policy-controller --devel --set cosign.secretKeyRef.name=mysecret
-```
-
-The secret specified should contain the key `cosign.pub` and the public key data content.
-
-Where `cosign.pub` is a file containing the public key, the kubernetes secret can be created with:
-```bash
-kubectl create secret generic mysecret -n cosign-system --from-file=cosign.pub=./cosign.pub
-```
-
-If the public key is able to validate a signature for the image digest, the admission controller will admit the image.
-If the public key is not able to validate a signature for the image digest, the admission controller will deny the image.
 
 ## Examples
 
--- a/pkg/webhook/validator.go
+++ b/pkg/webhook/validator.go
@@ -50,12 +50,12 @@ import (
 )
 
 type Validator struct {
-	client     kubernetes.Interface
+	client kubernetes.Interface
 }
 
 func NewValidator(ctx context.Context) *Validator {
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -47,10 +45,7 @@ An example of a denied admission would be:
 
 An example of no policy matched:
 1. If the image does not match against any policy
-1. Fallback to [deprecated policy-controller validation behavior](#deprecated-policy-controller-validation-behavior)
-1. Validation will be attempted against the secret defined under `cosign.secretKeyRef.name` during helm installation.
-  1. If a valid signature or attestation is obtained, image is admitted
-  1. If no valid signature or attestation is obtained, image is denied
+1. The image is admitted
```
vaikas, line 48: Maybe simplify these three lines to just say:
```
**If an image does not match a policy, it will be admitted**
```

Also, I think we should add a flag / env variable or some other configuration option. We might want to add another ConfigMap anyways (for adding upcoming multi TUF root work), so that might be something. I think having an option that says if an image does NOT match any policy should be denied should be there.
Maybe create an issue for it?

pkg/webhook/validator.go
```diff
@@ -288,14 +264,6 @@ func (v *Validator) validatePodSpec(ctx context.Context, namespace string, ps *c
 				continue
 			}
 			logging.FromContext(ctx).Errorf("ref: for %v", ref)
-			logging.FromContext(ctx).Errorf("container Keys: for %v", containerKeys)
-
-			if _, err := valid(ctx, ref, nil, containerKeys, ociremote.WithRemoteOptions(remote.WithAuthFromKeychain(kc))); err != nil {
```
vaikas, line 293: But if we do the above (add ability to configure what happens on 'no policy matches') as part of this PR, it might be cleaner to just do it now here, and we could use the 'FromContext' way to do this so that in this PR we configure this from the flag and later on we can change it to use ConfigMap if we so want to. Thoughts?

README.md
hectorj2f: > Maybe simplify these three lines to just say: **If an image does not match a policy, it will be admitted**

Sounds good to me.

> We might want to add another ConfigMap anyways (for adding upcoming multi TUF root work), so that might be something. 

👍🏻 

> I think having an option that says if an image does NOT match any policy should be denied should be there.

Why would we deny any image that doesn't match the policy ? If it doesn't match the globbing, then it should be excluded of the restrictions.

pkg/webhook/validator.go
hectorj2f: Perhaps I am missing something here. I understand the new logic as the following:

`policies, err := config.ImagePolicyConfig.GetMatchingPolicies(ref.Name())`

If `policies == 0` and `err == nil` the function should admit the image.
If `policies > 0` and `err != nil` the function should deny the image.

I think this is what it is happening now.

vaikas: Yes, but I'm just thinking that some folks might want to start by only requiring policies for a subset of images. I just recall there being interest in away to specify deny everything that doesn't match a policy. A safety catch all. Does that make sense?
By having a flag, would allow us to specify:
 * If an image doesn't match any policy allow (behaviour with this PR)
 * If an image doesn't match any policy deny (with the additional flag)

hectorj2f: I'm not so convinced unless you're aware of folks asking for this flag. We have added many fields to handle all type of validations, `mode` and `action`, so users can decide how strict to be. This flag sounds like another additional mechanism to enable additional logic.
 
>  I just recall there being interest in away to specify deny everything that doesn't match a policy. A safety catch all.

Wouldn't it be easier to create a CIP with `glob: '**'` to any of my authorities ? With this CIP, I force all the images to satisfy a specific authority, and so deny all. 

Perhaps I am not seen the use case this flag will cover.

vaikas: Because the CIPs are 'and'ed together, if you add a CIP with a ** => deny all, everything would be denied.
So to your point that for policies we have added ability to configure what would happen when a policy is met, that's facts, and it's very flexible. But for the case where an image does not match a policy, there's no way to specify what should happen today. With the existing flag the behaviour is for better or for worse that it will be denied because it will fail without signatures (uses Fulcio if no secret given as you know) and if a secret is given it's roughly the equivalent to '**' -> key to validate, but with one big exception, this '**' -> key does not get 'AND'ed with the other policies.
So, I'm just worried that as it exists today, the behaviour will be a breaking change on how things behave from a functional perspective.
How about this?
 * Put this in, and let's create an issue where we can discuss this with the broader community.
 * Let's socialize the issue in the sigstore policy-controller and gather some thoughts on it.

hectorj2f: My bad!. I omitted the 'and'ed current behaviour when having multiple CIPs. I now understand the purpose of having a flag.

Okay, let's follow the steps that you described above.

vaikas: LOL, I'm glad you were able to decode what I was trying to say even with formatting screwing things up, yikes :)

README.md
vaikas, line 48: Would you mind just updating this section to be bold and say something like:
```
If no policy matches an image, the image is admitted
```

so it's crystal clear :)

hectorj2f: Yes, I didn't pushed these changes. 
Done now!

README.md
```diff
@@ -45,12 +43,7 @@ An example of a denied admission would be:
 1. No valid signature or attestation was obtained for `policy2` with at least one of the `policy2` authorities
 1. The image is not admitted
 
-An example of no policy matched:
-1. If the image does not match against any policy
-1. Fallback to [deprecated policy-controller validation behavior](#deprecated-policy-controller-validation-behavior)
-1. Validation will be attempted against the secret defined under `cosign.secretKeyRef.name` during helm installation.
-  1. If a valid signature or attestation is obtained, image is admitted
-  1. If no valid signature or attestation is obtained, image is denied
+**If an image does not match a policy, it will be admitted**
```
cpanato, line 46: it will be or **not** be?
just a clarification question :)

vaikas: To be or not to be, that's the question :)

As it stands it WILL be admitted. Hence the #235 so that this behaviour can be controlled.

vaikas (approved):
So looks like other folks (besides me) think that having a default 'deny all' is useful so I think we should do that. As I mentioned in the bug, it could be a configmap, and with the TUF stuff we'll probably configure that through ConfigMap as well (ofc I could be wrong, but it seems like a good fit for it).
That's a bigger change, so let's get this in, and start adding support for the deny all.

README.md
hectorj2f: Yes, let's add support for that functionality.

Issue #234, opened 2022-09-13
Support non-default signature digest algorithms


joshrwolf:
**Description**

Plumb through `--signature-digest-algorithm` support to the CIP's spec to allow specifying non-sha256 signature digest algorithms.

This would be analogous to the `cosign` workflow of:

```bash
cosign verify --key $key --signature-digest-algorithm sha512 $image
```

hectorj2f:
Perhaps it could better fit as part of the CIP data model `authority.key.hashAlgorithm`  in https://github.com/sigstore/policy-controller/blob/main/pkg/apis/policy/v1beta1/clusterimagepolicy_types.go#L118

vaikas:
Yeah, I think it should definitely be plumbed through the CIP spec 👍

hectorj2f:
I am working on a branch to add support for this new field. I'll open PR soonish.

hectorj2f:
It has been added as part of in https://github.com/sigstore/policy-controller/pull/237

PR #237, opened 2022-09-15
Add a new field to set the signature hash algorithm

hectorj2f:
#### Summary

This PR adds a new field to the `authority.key` element setting the signature hash algorithm, e.g. sha256, sha512

```
kind: ClusterImagePolicy
metadata:
  name: demo
spec:
  images:
    - glob: '**'
  authorities:
  - name: simplekey
    key:
      hashAlgorithm: sha256
      data: |
        -----BEGIN PUBLIC KEY-----
        MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAE80swOaWDTfEDEqQoc7zWZWWM8xaZ
        TvYZCTJymLEl9hVgjBrCCoSbJ4QPncovTwyElYVuK1ajNDFovaensgl7QQ==
        -----END PUBLIC KEY-----
```

The supported hash algorithms are defined in https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/options/signature_digest.go#L30.

related to https://github.com/sigstore/policy-controller/issues/234

#### Documentation

I'll update our schema document with all the fields.

### DIFF
--- a/pkg/apis/policy/v1alpha1/clusterimagepolicy_types.go
+++ b/pkg/apis/policy/v1alpha1/clusterimagepolicy_types.go
@@ -127,7 +127,7 @@ type KeyRef struct {
 	// Supported formats differ based on the KMS system used.
 	// +optional
 	KMS string `json:"kms,omitempty"`
-	// HashAlgorithm always default to sha256 if the algorithm hasn't been explicitly set
+	// HashAlgorithm always defaults to sha256 if the algorithm hasn't been explicitly set
 	// +optional
 	HashAlgorithm string `json:"hashAlgorithm,omitempty"`
 }
--- a/pkg/apis/policy/v1alpha1/clusterimagepolicy_types.go
+++ b/pkg/apis/policy/v1alpha1/clusterimagepolicy_types.go
@@ -127,6 +127,9 @@ type KeyRef struct {
 	// Supported formats differ based on the KMS system used.
 	// +optional
 	KMS string `json:"kms,omitempty"`
+	// HashAlgorithm always default to sha256 if the algorithm hasn't been explicitly set
+	// +optional
+	HashAlgorithm string `json:"hashAlgorithm,omitempty"`
 }
 
 // StaticRef specifies that signatures / attestations are not validated but
--- a/pkg/apis/policy/v1beta1/clusterimagepolicy_types.go
+++ b/pkg/apis/policy/v1beta1/clusterimagepolicy_types.go
@@ -126,7 +126,7 @@ type KeyRef struct {
 	// Supported formats differ based on the KMS system used.
 	// +optional
 	KMS string `json:"kms,omitempty"`
-	// HashAlgorithm always default to sha256 if the algorithm hasn't been explicitly set
+	// HashAlgorithm always defaults to sha256 if the algorithm hasn't been explicitly set
 	// +optional
 	HashAlgorithm string `json:"hashAlgorithm,omitempty"`
 }
--- a/pkg/apis/policy/v1beta1/clusterimagepolicy_types.go
+++ b/pkg/apis/policy/v1beta1/clusterimagepolicy_types.go
@@ -126,6 +126,9 @@ type KeyRef struct {
 	// Supported formats differ based on the KMS system used.
 	// +optional
 	KMS string `json:"kms,omitempty"`
+	// HashAlgorithm always default to sha256 if the algorithm hasn't been explicitly set
+	// +optional
+	HashAlgorithm string `json:"hashAlgorithm,omitempty"`
 }
 
 // StaticRef specifies that signatures / attestations are not validated but
--- a/pkg/apis/policy/v1beta1/clusterimagepolicy_validation.go
+++ b/pkg/apis/policy/v1beta1/clusterimagepolicy_validation.go
@@ -23,6 +23,7 @@ import (
 	"strings"
 
 	"github.com/aws/aws-sdk-go-v2/aws/arn"
+	"github.com/sigstore/cosign/cmd/cosign/cli/options"
 	"github.com/sigstore/policy-controller/pkg/apis/glob"
 	"github.com/sigstore/sigstore/pkg/cryptoutils"
 	"k8s.io/apimachinery/pkg/util/sets"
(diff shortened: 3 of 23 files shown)


pkg/apis/policy/v1beta1/clusterimagepolicy_validation.go
```diff
@@ -23,6 +23,7 @@ import (
 	"strings"
 
 	"github.com/aws/aws-sdk-go-v2/aws/arn"
+	"github.com/sigstore/cosign/cmd/cosign/cli/options"
```
vaikas, line 26: I'd really really really not like to bring in cosign cli here.
Could we use things from here?
https://pkg.go.dev/crypto#pkg-constants

Or from sigstore/sigstore if they exist there?

hectorj2f: Yes, good call.

hectorj2f: I checked sigstore and there is not a clear list of support algorithm, as those identified in the cosign repo. So I decided to use our own validation, similar to the one in cosign here.

hectorj2f: Crypto package didn't have a function to validate the algorithm name, as those supported in cosign.

pkg/apis/policy/v1alpha1/clusterimagepolicy_types.go
```diff
@@ -127,6 +127,9 @@ type KeyRef struct {
 	// Supported formats differ based on the KMS system used.
 	// +optional
 	KMS string `json:"kms,omitempty"`
+	// HashAlgorithm always default to sha256 if the algorithm hasn't been explicitly set
```
vaikas, line 130: nit s/default/defaults/

pkg/apis/policy/v1beta1/clusterimagepolicy_types.go
```diff
@@ -126,6 +126,9 @@ type KeyRef struct {
 	// Supported formats differ based on the KMS system used.
 	// +optional
 	KMS string `json:"kms,omitempty"`
+	// HashAlgorithm always default to sha256 if the algorithm hasn't been explicitly set
```
vaikas, line 129: s/default/defaults/

pkg/apis/signaturealgo/signature_digest.go
```diff
@@ -0,0 +1,48 @@
...
+	"sha256": crypto.SHA256,
+	"sha384": crypto.SHA384,
+	"sha512": crypto.SHA512,
+}
+
+// HashAlgorithm returns a crypto.Hash code using an algorithm name as input parameter
+func HashAlgorithm(algorithmName string) (crypto.Hash, error) {
+	normalizedAlgo := strings.ToLower(strings.TrimSpace(algorithmName))
+
+	if normalizedAlgo == "" {
```
vaikas, line 38: nit, maybe just move this above normalizing so if it's empty, we don't ? Though I guess if somebody passes in " " then it would still work this way.

pkg/webhook/clusterimagepolicy/clusterimagepolicy_types.go
```diff
@@ -254,9 +271,19 @@ func convertKeyRefV1Alpha1ToWebhook(in *v1alpha1.KeyRef) *KeyRef {
 	if in == nil {
 		return nil
 	}
+	// Convert the hash algorithm name to the code and reuse it evertwhere else
```
vaikas, line 274: nit, everywhere

pkg/apis/signaturealgo/signature_digest_test.go
```diff
@@ -0,0 +1,45 @@
...
+// limitations under the License.
+
+package signaturealgo
+
+import (
+	"crypto"
+	"testing"
+)
+
+func TestHashAlgorithm(t *testing.T) {
```
vaikas, line 22: nit, maybe add an invalid test case. I see it tested other places, but might as well test it here to get full coverage in this unit test.

pkg/webhook/clusterimagepolicy/clusterimagepolicy_types.go
```diff
@@ -80,6 +81,12 @@ type KeyRef struct {
 	// Data contains the inline public key
 	// +optional
 	Data string `json:"data,omitempty"`
+	// HashAlgorithm always default to sha256 if the algorithm hasn't been explicitly set
```
vaikas, line 84: I was wondering why we need both of these. I assume it's because we can't serialize HashAlgorithmCode below? If that's the case, maybe add a comment here (just like we do for PublicKeys below).

vaikas (commented):
Looks great, thanks!
Would you mind adding an e2e test with a different hash algorithm?

pkg/webhook/clusterimagepolicy/clusterimagepolicy_types.go
hectorj2f: Yes, I wanted to serialize the algo name in the configmap while having in the object the code so we can directly use it.

pkg/apis/signaturealgo/signature_digest.go
hectorj2f: i tested a failed scenario here `{algorithm: "sha3845", wantErr: true, wantHash: crypto.SHA256},`

test/e2e_test_cluster_image_policy.sh
```diff
@@ -435,3 +435,31 @@ kubectl delete cip --all
...
+echo '::endgroup::'
+
+echo '::group:: Deploy ClusterImagePolicy With signature digest sha512'
+yq '. | .metadata.name = "image-policy-sha512-key"' \
+  ./test/testdata/policy-controller/e2e/cip-key-hash-algorithm.yaml | \
+  kubectl apply -f -
+echo '::endgroup::'
+
+echo '::group:: Verify ClusterImagePolicy With signature digest sha512 using a pod'
+# We signed this above and applied remote signature source location above
```
vaikas, line 453: super nit: We didn't sign this, it's already signed by the image authors.

vaikas (approved):
We can fix the nit later in some other pr as a cleanup if you prefer that :)

vaikas (approved):
Thank you!!!

hectorj2f:
@vaikas  thanks for the review 💪🏻

Issue #138, opened 2022-08-08
Add support for non-public, and multiple TUF roots and a way to configure them.


vaikas:
**Description**
Currently there's no way to use different TUF roots. This is a problem if for example you run a private instance of Sigstore and want to still use Public instance of Sigstore. We currently (in Sigstore as a whole) have some environmental variables to configure to trust different custom Sigstore instances, but we're working on getting rid of those and only use TUF for establishing trust.

There's a proposal here:
https://docs.google.com/document/d/1QWBvpwYxOy9njAmd8vpizNQpPti9rd5ugVhji0r3T4c/edit

That (I think) will require changes to the API surface. The user will need to specify for keyless entries which root should be used, and therefore that will have to be a new field in the API.

Furthermore, adding new roots must be documented, and one option is to mount the different roots into the container from secret that contains the tuf root information into the right place. The one wrinkle however is that what's being mounted is a directory tree and not a single (or couple of files). One option here might be to tar up the tree into a secret and upon restart untar the tree to the appropriate location.

Anyways, this issue tracks that work as an umbrella issue.

vaikas:
Ok, #157 Adds support for initializing TUF for different root. But, it does not allow for multiple ones yet, so will need to add support for that both in the API as well as how things get initialized.

vaikas:
https://github.com/sigstore/sigstore/pull/615

If my understanding is correct we should not have to specify the TUF at the API level as long as the targets get constructed properly. I reckon we might be able to map fulcio/rekor etc. URLs so that what gets put in the API allows us to the the appropriate TUF entries from multiple roots. I assume we'd use ConfigMap to configure multiple roots so that it can then be slurped up into the policy-controller.

vaikas:
Ok, here's a short document that I'd like to get some feedback on. Since I started the investation on how this may look the sigstore proto buf work has started in earnest, so using that also to align hopefully these two efforts.

https://docs.google.com/document/d/1a-tV9YyXHvtoaZXm-eMcqP-zFTWg6Pqfjyl8W4uS9F4/edit?usp=sharing

Alignment is basically around this struct here for users wanting to bring their own keys independent of TUF as a delivery mechanism for trusted keys.
https://github.com/sigstore/protobuf-specs/pull/5/files#diff-b1f89b7fd3eb27b519380b092a2416f893a96fbba3f8c90cfa767e7687383ad4R70

And lastly there's some prototyping here:
#291

JAORMX:
[...]

PR #240, opened 2022-09-16
Rely exclusively on TUF root for fulcio root. Do not fetch them oob.

vaikas:

#### Summary
Currently in order to support different instance of Fulcio, we fetch the public root cert from the Fulcio. This is not ideal for many reasons, but was a work around until we had a way to support non-standard TUF root for testing. Now that we're moving to TUF and we have a way to test with non-standard TUF, remove the last bit of that functionality.

We still have to solve this:
https://github.com/sigstore/policy-controller/issues/138

#### Release Note
Do not fetch Fulcio root cert from the Fulcio API and rely on TUF root instead.


cpanato (approved):

hectorj2f (commented):
@vaikas If we don't consume the `keyless.url` value anymore. I was wondering if we could add a validation to only allow `keyless.url=https://fulcio.sigstore.dev`.  It is a configurable field in the CIP today.

hectorj2f (approved):
lgtm

vaikas:
We do still use it, but right now it's implicitly tied to the TUF root value. Once we support more than one TUF root this will be needed to select between which TUF root to use.

PR #248, opened 2022-09-21
Feature: match available resource types by name, version, group and/or labels

hectorj2f:
#### Summary

This PR adds a new field `match` to the CIP. This new field sets the type of core resources, and a list of labels (optional) that must match specific resources in the cluster. In other words, these constraints define which resources will be enforced by the CIP. 
This behavior differs from the current logic where all core resource types are enforced within a namespace.

For now, the policy-controller still requires users to label the namespaces (sigstore/include=true) to enforce the existing CIPs.

```
Spec:
  # Once the namespace has been labeled (sigstore/include=true), this    
  # new field `match.any` allows     
  # defining which kind of resource core types the policy will  
  # enforce. 
  # Likewise this field also allows selecting resources based     
  # on their name following regular expressions. If resources.kinds 
  # are not defined, the old default behavior will apply to all the 
  # core resource types.
  match:
      # This list of resources allows to choose which resource to 
      # enforce based on the group, version, name or by using labels.   
      # If the list of matching resources is empty, it defaults to
      # all core resources, as done in the past, replicaset, cronjob
      # job, deployment, pod, daemonset, statefulset.
      - group: batch
        resource: cronjobs
        version: v1beta1
        # selects a resource type based on the existence of labels
        selector:
           matchLabels:
chainguard: enforce
environment: prod
        
      # kinds defines a list of core resource types for which the 
      # policy will be enforced.
 - version: v1
   resource: pods
      - group: apps
   version: v1
   resource: deployments
   ```
   
The following are examples of CIPs where we use this new field `match`:

```
apiVersion: cosigned.sigstore.dev/v1beta1
kind: ClusterImagePolicy
metadata:
  name: image-policy
spec:
  match:
  -  resource: jobs
     group: batch
      version: v1
  -  resource: pods
      version: v1
  images:
  - glob: *
  authorities:
  - keyless:
      url: https://fulcio.mattmoor.dev/
      identities:
      - issuer: https://container.googleapis.com/v1/projects/myco-prod/locations/*/clusters/*
        subject: https://k8s.io/namespaces/*/serviceaccounts/*

    tlog:
      url: https://rekor.mattmoor.dev
```
or another example using labels:

```
apiVersion: cosigned.sigstore.dev/v1beta1
kind: ClusterImagePolicy
metadata:
  name: image-cronjob-policy
spec:
  match:
      - resource: cronjobs
         group: batch
         version: v1
         selector:
           matchLabels:
              prod: x-cluster1
  images:
  - glob: *
  authorities:
  - keyless:
      url: https://fulcio.mattmoor.dev/
      identities:
      - issuer: https://container.googleapis.com/v1/projects/myco-prod/locations/*/clusters/*
        subject: https://k8s.io/namespaces/*/serviceaccounts/*

    tlog:
      url: https://rekor.mattmoor.dev
```

related to: https://github.com/sigstore/policy-controller/issues/18 
related to: https://github.com/chainguard-dev/mono/issues/1935

### CODE REVIEW

**Comment by hectorj2f** [comment]
I am re-running the failed jobs, as I got unrelated errors while running:

```
curl -fsL https://github.com/ko-build/ko/releases/download/null/ko_ull_Linux_x86_64.tar.gz

gzip: stdin: unexpected end of file
tar: Child returned status 1
```

File: pkg/apis/config/image_policies.go [hunk 1]
```diff
@@ -76,10 +118,69 @@ func parseEntry(entry string, out interface{}) error {
...
+	}
+	return ret, lastError
+}
+
+// GetMatchingPolicies returns all matching Policies and their Authorities that
+// need to be matched the Image.
 // Returned map contains the name of the CIP as the key, and a normalized
 // ClusterImagePolicy for it.
-func (p *ImagePolicyConfig) GetMatchingPolicies(image string) (map[string]webhookcip.ClusterImagePolicy, error) {
+func (p *ImagePolicyConfig) GetMatchingPoliciesByImage(image string) (map[string]webhookcip.ClusterImagePolicy, error) {
```
vaikas, line 183: Is this now only being used by tester?

test/testdata/policy-controller/e2e/cip-match-resource-label.yaml
```diff
@@ -0,0 +1,36 @@
...
+apiVersion: policy.sigstore.dev/v1alpha1
+kind: ClusterImagePolicy
+metadata:
+  name: image-policy-match-label
+spec:
+  images:
+  - glob: "**demo**"
+  match:
+  - resource: pods
+    version: v1
```
vaikas, line 24: I wonder if we should have a '*' for these, for things like cronjobs? Or maybe I guess you could for cronjobs then have to add both of them (v1, v1beta1?), but then we'd have to support it in the lookup table, which now just has v1 there.
WDYT?

pkg/apis/config/image_policies.go
hectorj2f: yes, that is correct. I wanted to leave it, but I am happy to polish this code.

test/testdata/policy-controller/e2e/cip-match-resource-label.yaml
hectorj2f: I think we can use '*', and add the missing versions to the supported api types. That looks more robust in my opinion. In the future we could add a checking in the policy validating webhook, so the types are not hardcoded in the webhook itself.

pkg/apis/config/image_policies.go
vaikas: I guess my only thinking is that if they then start diverging. Maybe just get rid of it now, and pass in fields in there that don't restrict anything?

hectorj2f: Sounds good to me!

vaikas (approved):

Issue #235, opened 2022-09-13
Should the controller deny an image if no policy matches  ?


hectorj2f:
**Question**

As a result of a discussion in https://github.com/sigstore/policy-controller/pull/223 and due to the AND'ed nature of the CIPs, we need to understand any potential use cases whenever an image does not match any of the policies. 

The current implementation (added in https://github.com/sigstore/policy-controller/pull/223) admits images not matching any of the list of available policies. However there could be scenarios whenever users might want to deny all the images that don't match any of the policies.

vaikas:
I too think it's useful (as I said in the code review). Haven't heard anybody have any opposing views, so I reckon we should make this an option. I think we probably are going to need a configmap for configuring different TUF roots (don't want them getting mounted, etc.) tracked in #138. So, maybe this is as good of a time as any to add PolicyController configmap so that it can be configured there?
It's a bigger change, so I reckon we get #223 in, then add the configmap support for configuring this deny all behaviour.

vaikas:
What should the default be here in the configmap?
https://github.com/sigstore/policy-controller/pull/270
@DennyHoang 
@elfotografo007 
@hectorj2f 
@tysonkamp (this wouldn't autocomplete for me, so I hope I got it right? 😰 )

Just tagging y'all, since you expressed thumbs up here. Want to get this sorted before the next release. In the #270 the default config in the configmap is allow, but easy enough to change now rather than change after the release.

hectorj2f:
By default, I'd say `deny` if there isn't any matching policy. It is the right way due to security reasons. It is also aligned with the old behaviour, as we were checking the signatures against fulcio by default.

tysonkamp:
I gave the thumbs up above since absent other potentialy nuances here the idea of denying by defatul follows the principle of least privilege. I can think of examples in other domains where this wasn't followed and by the time the vendor made "secure by default" the actual default, they got so much push back from customers because it broke  functionality in their apps that the vendor had to roll it back and "insecure by default" was the final state of things.  The instance I'm referring to opened apps up for HTTP request smuggling :/

mattmoor:
I think aligning with the old behavior is the most user friendly approach here.  Folks still need to label namespaces, so it'd be great to avoid exposing gaps by default.

elfotografo007:
I also support the `deny` default.

vaikas:
I also believe deny by default is the right behaviour 👍 
[...]

PR #270, opened 2022-09-26
Add policy-controller level config map and decorate requests with it.

vaikas:

#### Summary
Start putting the pieces together for allowing configuration for:
#235


config/kustomization.yaml.nope
```diff
@@ -0,0 +1,32 @@
+# Copyright 2022 The Sigstore Authors.
```
hectorj2f, line 1: why do we need this file ?

go.mod
```diff
@@ -2,6 +2,13 @@ module github.com/sigstore/policy-controller
 
 go 1.18
 
+// Bring this in for testing:
```
hectorj2f, line 5: Should we skip adding this ? I assume we don't need it for this code. Right ?

pkg/config/store.go
```diff
@@ -0,0 +1,134 @@
...
+// Point is that these apply to the whole controller instead of specific CIP
+// policies that apply only to matching images.
+type PolicyControllerConfig struct {
+	// NoMatchPolicy says what do in the case where an image does not match
+	// any policy.
+	NoMatchPolicy string `json:"no-match-policy"`
+}
+
+func NewPolicyControllerConfigFromMap(data map[string]string) (*PolicyControllerConfig, error) {
+	ret := &PolicyControllerConfig{NoMatchPolicy: "deny"}
```
hectorj2f, line 57: The default behavior should be `admit`, right ? At least, the configmap sets the default to `admit`.

pkg/config/store.go
```diff
@@ -0,0 +1,134 @@
...
+
+// FromContextOrDefaults is like FromContext, but when no
+// PolicyControllerConfig is attached, it returns a PolicyControllerConfig
+// populated with the defaults for each of the fields.
+func FromContextOrDefaults(ctx context.Context) *PolicyControllerConfig {
+	if cfg := FromContext(ctx); cfg != nil {
+		return cfg
+	}
+	return &PolicyControllerConfig{
+		NoMatchPolicy: DenyAll,
```
hectorj2f, line 92: Generally I'd vote for `deny` everything that doesn't match the policies, but I learnt that a blocking by-default behaviour might be too strict for new users.

hectorj2f (commented):
lgtm, so far. I just made few minor comments.

pkg/webhook/validator.go
```diff
@@ -254,13 +255,31 @@ func (v *Validator) validatePodSpec(ctx context.Context, namespace string, ps *c
...
-				logging.FromContext(ctx).Errorf("policies: for %v", policies)
 			}
 
 			if passedPolicyChecks {
 				logging.FromContext(ctx).Debugf("Found at least one matching policy and it was validated for %s", ref.Name())
 				continue
 			}
+			// There were no matching policies found against this image.
+			// Check what the configuration is and act accordingly.
+			pcConfig := policycontrollerconfig.FromContext(ctx)
```
hectorj2f, line 266: nit: I'd move this code to its own function, this `validatePodSpec` is getting big.

pkg/config/store.go
vaikas: Yeah, that's why I made the configmap by default allow all to match current behaviour. However in case of any mistakes with the configmap, I want to fail closed by default..

pkg/config/store.go
vaikas: Yes, that's the default behaviour out of the box. But, in case of errors, I believe doing the secure thing by default is correct thing to do.

go.mod
vaikas: haha, yes, was testing something else and these got committed by mistake, thanks for catching this!

pkg/webhook/validator.go
vaikas: done

config/kustomization.yaml.nope
vaikas: haha, got added to commit by mistake, good catch :)

pkg/config/store.go
hectorj2f: 👍🏻

pkg/config/store.go
hectorj2f: We'll need to document this behaviour.

hectorj2f (approved):

mattmoor:
Why does this introduce another store (`pkg/config`) instead of adding to the existing store (`pkg/apis/config`)?

You should be able to put as many configmaps into each store as you want.

vaikas:
Yeah, I thought it would be cleaner having it be totes separate, but I can
add it to the existing one.

PR #279, opened 2022-09-29
deny by default.

### LINKED ISSUE #235: Should the controller deny an image if no policy matches  ?

vaikas:

#### Summary
fixed #235 

#### Release Note
* Any image not matching a policy is 'deny' by default.

### DIFF
--- a/config/config-policy-controller.yaml
+++ b/config/config-policy-controller.yaml
@@ -18,4 +18,4 @@ metadata:
   name: config-policy-controller
   namespace: cosign-system
 data:
-  no-match-policy: allow
+  no-match-policy: deny
--- a/test/e2e_test_cluster_image_policy.sh
+++ b/test/e2e_test_cluster_image_policy.sh
@@ -458,12 +458,26 @@ yq '. | .metadata.name = "image-policy-match-label"
 sleep 5
 echo '::endgroup::'
 
+# For pods that do not match labels, meaning there are no matching policies, we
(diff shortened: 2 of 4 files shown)


mattmoor (approved):

k4leung4 (approved):
thanks for the fix.

hectorj2f (approved):
lgtm to deny-all

config/config-policy-controller.yaml
```diff
@@ -18,4 +18,4 @@ metadata:
   name: config-policy-controller
   namespace: cosign-system
 data:
-  no-match-policy: allow
+  no-match-policy: deny
```
mattmoor, line 21: If this is the default, then I'd recommend that we omit it entirely.

By specifying this, we make it so that folks upgrading using our configs will overwrite any changes they make to these.

hectorj2f: But this configmap hasn't been released yet. I don't see any harm on doing this change to the config file.

hectorj2f: Okay, i understood the intention now.

Issue #278, opened 2022-09-29
Improve behavior for ClusterImagePolicy with multiple keys


wojciechka:
**Description**

When providing a `ClusterImagePolicy` with multiple (N) keys in `spec.authorities`, `policy-controller` loops over all keys, calling `cosign` package to perform verification, which in turn pulls the signature image N * M times, where `M` is the number of signatures in OCI registry (and `N` is the number of public keys in `ClusterImagePolicy`).

https://github.com/sigstore/policy-controller/blob/v0.3.0/pkg/webhook/validator.go#L610-L614
https://github.com/sigstore/policy-controller/blob/v0.3.0/pkg/webhook/validation.go#L47-L64 - looping over keys
https://github.com/sigstore/policy-controller/blob/v0.3.0/pkg/webhook/validation.go#L73-L81 - calling `cosign.VerifyImageSignatures`

The `cosign.VerifyImageSignatures` call pulls image, which in turn means that having N signatures is calling the image registry N times (or until a valid signature is found).

So, with 15 keys and 3 signatures attached to an image, this means the pull will be done 45 (15 * 3) times.

One option to improve is to download the signature to a local file and use `cosign.VerifyLocalImageSignatures` which is not doing any networking and should be much faster.

https://github.com/sigstore/policy-controller/blob/v0.3.0/pkg/webhook/validator.go#L610-L614 also mentioned a comment about what to do with multiple keys, I am not sure if this should be same or another issue, but I'll add it here as it may be related.

I suggest this defaults to any key working as it is today, but ideally, `spec.authorities` could have a field like `threshold` or `minimumSignatures` that would specify minimum number of signatures that have to be present - for example 3 out of 5. This would be similar to the `threshold` option in [connaisseur](https://github.com/sse-secure-systems/connaisseur), as documented [here](https://sse-secure-systems.github.io/connaisseur/v2.6.0/validators/sigstore_cosign/#multi-signature-verification).

I am using and checked the source code for version `v0.3.0`.

hectorj2f:
@wojciechka We are working on improvement the performance overall for the controller. I agree we could improve that piece of code to be more effecting when pulling the signature, but it also depends on the type of algorithm of the public key (e.g. sha512). Maybe we could split that function in a way where the signature is pulled before creating the verifier. 
On the other hand, we are pulling the manifest, not the image.

wojciechka:
> (...) but it also depends on the type of algorithm of the public key (e.g. sha512).

[...]

PR #283, opened 2022-10-01
Cache scaffolding

vaikas:
#### Summary
Start wiring up caching as per #278 that caches Policies against an image and more importantly against a particular version of a CIP. Need to make sure we're checking against the latest and greatest of the CIP, so add fields for those. At the moment doesn't do anything useful, as far as caching goes, but starting to see how it could look like.


cpanato (approved):
Cool

hectorj2f (approved):
lgtm

PR #284, opened 2022-10-01
See what breaks after adding insecure registry test back.

vaikas:

#### Summary
Follow up comment investigation to here:
https://github.com/sigstore/policy-controller/commit/7e457bb59aaa8d6d1f4ad7ff9d52cab7b1b81db5#commitcomment-85428835

### DIFF
--- a/.github/workflows/kind-e2e-cosigned.yaml
+++ b/.github/workflows/kind-e2e-cosigned.yaml
@@ -112,6 +112,11 @@ jobs:
         kubectl rollout status --timeout 5m --namespace cosign-system deployments/webhook
         kubectl wait deployment -n cosign-system --for condition=Available=True --timeout=90s --all
 
+    - name: Run Insecure Registry Tests
+      run: |
+        go install github.com/google/go-containerregistry/cmd/crane
+        ./test/e2e_test_insecure_registry.sh
+
     - name: Run Image Policy Tests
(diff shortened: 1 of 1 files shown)


.github/workflows/kind-e2e-cosigned.yaml
```diff
@@ -112,6 +112,11 @@ jobs:
         kubectl rollout status --timeout 5m --namespace cosign-system deployments/webhook
         kubectl wait deployment -n cosign-system --for condition=Available=True --timeout=90s --all
 
+    - name: Run Insecure Registry Tests
+      run: |
+        go install github.com/google/go-containerregistry/cmd/crane
```
cpanato, line 117: maybe we can use `  - uses: imjasonh/setup-crane@v0.2` to setup crane

cpanato (approved):
non blocking comment

.github/workflows/kind-e2e-cosigned.yaml
hectorj2f: +1

vaikas:
It really should be part of the cosign test suite and was removed here in the repository shuffle.
https://github.com/sigstore/cosign/blob/main/test/e2e_test_insecure_registry.sh
I'm closing this one and we can add it there instead.

Issue #268, opened 2022-09-25
Validate ephemeral containers also


vaikas:
**Description**
Previous discussion here, moving to this repo :) 
https://github.com/sigstore/cosign/issues/820

PR #299, opened 2022-10-11
Feature: add support for ephemeral containers

hectorj2f:

#### Summary

I am working on adding support for ephemeral pods in the webhook validation.

closes https://github.com/sigstore/policy-controller/issues/268

I tested my changes using the following examples:

```
kubectl create namespace test

kubectl label ns test policy.sigstore.dev/include=true --overwrite

kubectl run mynginx -n test --image=cgr.dev/chainguard/nginx

kubectl debug -it mynginx -n test --image=busybox@sha256:9810966b5f712084ea05bf28fc8ba2c8fb110baa2531a10e2da52c1efc504698
Defaulting debug container name to debugger-drfgz.
Error from server (BadRequest): admission webhook "policy.sigstore.dev" denied the request: validation failed: failed policy: demo: spec.ephemeralContainers[0].image
busybox@sha256:9810966b5f712084ea05bf28fc8ba2c8fb110baa2531a10e2da52c1efc504698 signature keyless validation failed for authority authority-0 for index.docker.io/library/busybox@sha256:9810966b5f712084ea05bf28fc8ba2c8fb110baa2531a10e2da52c1efc504698: no matching signatures:
```

I used a dummy CIP that validated any image against the remote sigstore.dev rekor.

#### Release Note

Add support for Ephemeral Containers.


hectorj2f:
I am looking why my validating webhook configuration is not been respected even if I am using https://github.com/knative/pkg/pull/2547.

vaikas:
You mean for the ephemeral containers, or just in general?

hectorj2f:
Yes, I mean for the ephemeral containers, once I check the configuration in the cluster. I don't see the `pods/ephemeralcontainers` anywhere in my configuration.

config/500-webhook-configuration.yaml
```diff
@@ -32,7 +32,17 @@ webhooks:
   failurePolicy: Fail
   sideEffects: None
   timeoutSeconds: 25
-
+  rules:
```
vaikas, line 35: These get overwritten, so these won't have any effect. You need to add these here:
https://github.com/sigstore/policy-controller/blob/main/cmd/webhook/main.go#L125

And might have to add a new one for Pod that supports the `ephemeralcontainers` subresource here:
https://github.com/sigstore/policy-controller/blob/main/cmd/webhook/main.go#L112

hectorj2f: gotcha! Thanks for the feedback @vaikas :)

hectorj2f: Fixed now 🎉 !

hectorj2f:
@vaikas This PR is ready for a review 🎉

cmd/webhook/main.go
```diff
@@ -103,26 +103,44 @@ func main() {
...
+
+	_ resourcesemantics.SubResourceLimited = (*crdEphemeralContainersUpdatesOrDeletes)(nil)
+	_ resourcesemantics.VerbLimited        = (*crdEphemeralContainersUpdatesOrDeletes)(nil)
 )
 
 type crdNoStatusUpdatesOrDeletes struct {
 	resourcesemantics.GenericCRD
 }
 
+type crdEphemeralContainersUpdatesOrDeletes struct {
```
vaikas, line 115: Bit of a nit. I'd prefer this to be just:
crdEphemeralContainers

Previously it had the NoStatusUpdatesOrDeletes, and now that it only pertains to EphemeralContainers and it doesn't have the 'No' there it's a bit misleading cause I read it as it's only for Updates Or Deletes for EphemeralContainers :)

pkg/webhook/validator.go
```diff
@@ -846,8 +912,8 @@ func getNamespace(ctx context.Context, namespace string) string {
 // All the matched policies were validated, or
 // no matching policies were found, but the PolicyControllerConfig has been
 // configured to allow images not matching any policies.
-func (v *Validator) validateContainer(ctx context.Context, c corev1.Container, namespace, field string, index int, kind, apiVersion string, labels map[string]string, ociRemoteOpts ...ociremote.Option) *apis.FieldError {
-	ref, err := name.ParseReference(c.Image)
+func (v *Validator) validateContainer(ctx context.Context, containerImage string, namespace, field string, index int, kind, apiVersion string, labels map[string]string, ociRemoteOpts ...ociremote.Option) *apis.FieldError {
```
vaikas, line 915: nit, wonder if we should now rename this (internal so should be fine) to:
validateContainerImage, or just validateImage?

test/e2e_test_cluster_image_policy.sh
```diff
@@ -544,3 +544,87 @@ echo '::group::' Cleanup
...
+  --namespace cosign-system \
+  --type merge \
+  --patch '{"data":{"no-match-policy":"deny"}}'
+# allow for propagation
+sleep 10
+echo '::endgroup::'
+
+ephemeralContainerImage="busybox@sha256:9810966b5f712084ea05bf28fc8ba2c8fb110baa2531a10e2da52c1efc504698"
+
+echo '::group:: test rejection of ephemeral container that does not have any signature'
```
vaikas, line 617: Could we add a passing test case for the ephemeral container also? L598 creates the pod, but we don't seem to be (or I missed it), test the passing case?

Maybe just add:

```
if ! kubectl debug poddemo -n ${NS} --image=$(demoEphemeralContainerImage} ; then
  echo Failed to create a signed ephemeralcontainer ...
```

test/e2e_test_cluster_image_policy.sh
```diff
@@ -544,3 +544,87 @@ echo '::group::' Cleanup
...
+go mod init example.com/demo
+cat <<EOF > main.go
+package main
+import (
+  "fmt"
+  "time"
+)
+func main() {
+  // Calling Sleep method
+  time.Sleep(8 * time.Minute)
```
vaikas, line 560: No action necessary, just jotting this down :)

Was worried that this pod would hang around for 8 minutes, but looking at runtimes for the tests that's not the case  😰 🤣

vaikas (commented):
Thanks for doing this!!!

test/e2e_test_cluster_image_policy.sh
hectorj2f: yes

pkg/webhook/validator.go
hectorj2f: I agree, somehow i don't know why i missed to rename it. good catch

cmd/webhook/main.go
hectorj2f: 👍🏻

test/e2e_test_cluster_image_policy.sh
hectorj2f: The intention is to have a running pod, so we can create an ephemeral one.

hectorj2f: I could reduce the sleep time to 3min or so, i didn't want to be pick with the time

test/e2e_test_cluster_image_policy.sh
hectorj2f: This action reported an error when running Kubernetes 1.22, it makes sense as ephemeral containers were enabled by default in 1.23.

vaikas (approved):
Awesomesauce!!! Thank you!!!

PR #303, opened 2022-10-11
Fix: Check flag in v1alpha1, use OrDefaults to avoid breaking change.

mattmoor:
:bug: The addition of the new config to v1beta1 without `OrDefaults` is breaking (segfault!) to downstream callers of `Validate()`, so this changes things to use `OrDefaults` for when that configuration is unavailable.
:bug: I also noticed that the v1alpha1 validation wasn't checking the config, so depending on the version we validate with this is breaking, which creates some weird semantics with conversion.

This makes things use `OrDefaults` and copies the logic to `v1alpha1`.

/kind bug


mattmoor:
cc @vaikas @hectorj2f

hectorj2f (approved):

vaikas (approved):

PR #304, opened 2022-10-11
update README with some new features

hectorj2f:

#### Summary

There were some major changes to the README that we had to include.


cpanato (approved):
🚀

vaikas (approved):
Thank you!

Wonder if we should specify what the default behaviour is? 'deny'?
Also, can you file an issue (and feel free to assign to me) but we need to update the sigstore.dev docs for all these new features and how to configure them.

PR #328, opened 2022-10-18
update images to the new path

cpanato:
#### Summary
- update images to the new path

there are some tests that is using the old path, i will try to update those, did not made in this PR because they are failing and I will need more time to figure out


vaikas (approved):
Thanks!!

PR #337, opened 2022-10-21
Add validation for the oci repository field

hectorj2f:
#### Summary

Add a validation to avoid setting wrong values for the source.OCI.

#### Release Note

Add a validation  to avoid setting wrong values for the source.OCI.

### DIFF
--- a/pkg/apis/policy/v1alpha1/clusterimagepolicy_validation.go
+++ b/pkg/apis/policy/v1alpha1/clusterimagepolicy_validation.go
@@ -18,12 +18,12 @@ import (
 	"context"
 	"fmt"
 	"net"
-	"net/url"
 	"path/filepath"
 	"regexp"
 	"strings"
 
 	"github.com/aws/aws-sdk-go-v2/aws/arn"
+	registryfuncs "github.com/google/go-containerregistry/pkg/name"
 	"github.com/sigstore/policy-controller/pkg/apis/glob"
 	"github.com/sigstore/policy-controller/pkg/apis/signaturealgo"
 	"github.com/sigstore/sigstore/pkg/cryptoutils"
@@ -34,9 +34,8 @@ import (
 )
 
 const (
-	awsKMSPrefix       = "awskms://"
-	ociRepoDelimiter   = "/"
-	ociRepositoryChars = "abcdefghijklmnopqrstuvwxyz0123456789_-./"
+	awsKMSPrefix     = "awskms://"
+	ociRepoDelimiter = "/"
 )
 
 var (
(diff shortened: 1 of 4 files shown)


pkg/apis/policy/v1beta1/clusterimagepolicy_validation_test.go
```diff
@@ -177,6 +177,64 @@ func TestKeyValidation(t *testing.T) {
...
+		{
+			name:        "Should fail with invalid OCI value",
+			errorString: "invalid value: registry.example.com/repo/*: spec.authorities[0].source[0].oci\ncan only contain the characters `abcdefghijklmnopqrstuvwxyz0123456789_-./`",
+			policy: ClusterImagePolicy{
+				Spec: ClusterImagePolicySpec{
+					Images: []ImagePattern{{Glob: "gcr.io/*"}},
+					Authorities: []Authority{
+						{
+							Key:     &KeyRef{KMS: "kms://key/path"},
+							Sources: []Source{{OCI: "registry.example.com/repo/*"}},
```
mattmoor, line 189: Check a DockerHub style source without a hostname?

e.g. `mattmoor/signatures` should probably work :)

hectorj2f: yes, `mattmoor/signatures `will work, but `mattmoor` could be a valid hostname here.

pkg/apis/policy/v1alpha1/clusterimagepolicy_validation.go
```diff
@@ -218,6 +223,36 @@ func (source *Source) Validate(ctx context.Context) *apis.FieldError {
 	var errs *apis.FieldError
 	if source.OCI == "" {
 		errs = errs.Also(apis.ErrMissingField("oci"))
+	} else {
+		var registry string
```
vaikas, line 227: Instead of having this duplicated in both validation functions, could we hoist this into a function somewhere else? Also, side note, I was surprised from a quick googling that there's not something that we could use for this.

mattmoor: Maybe: https://github.com/google/go-containerregistry/blob/02f47e132cf562183cfd7cf272c0605e6183ba53/pkg/name/repository.go#L71-L72

vaikas: There we go, I poked there but couldn't find it, but looks like that may do it?

hectorj2f: Yes, I reused the validations from the repository and registry names validations (which are private func). I assume we can accept a registry name and/or a repo name.

hectorj2f: Do you have any contrary thoughts about it? Otherwise I can move on with this function and move it to a generic place. Or do you want me to adapt the functions in go-containerregistry to fit our needs?

mattmoor: What I was thinking was just calling:
```go
_, err := name.NewRepository(foo)
if err != nil {
   return &apis.FieldError{
      Message: err.Error(),
   }
}
```

This is what I've done in the past to access this validation logic.

hectorj2f: I don't like the function NewRepository() because it only focuses about a repository name, but I would say we need to validate a registry name (if that is what they only provided) e.g. `us.gcr.io`. 


I opened this PR:  https://github.com/google/go-containerregistry/pull/1474

hectorj2f: I closed this PR  https://github.com/google/go-containerregistry/pull/1474, so this can get fixed asap. I adjusted the code to accept both registry/repo and registry.

mattmoor (approved):
cc @imjasonh

pkg/apis/policy/v1alpha1/clusterimagepolicy_validation.go
```diff
@@ -218,6 +222,20 @@ func (source *Source) Validate(ctx context.Context) *apis.FieldError {
 	var errs *apis.FieldError
 	if source.OCI == "" {
 		errs = errs.Also(apis.ErrMissingField("oci"))
+	} else {
```
imjasonh, line 225: WDYT about the suggestion here: https://github.com/google/go-containerregistry/pull/1474#issuecomment-1289024708 

This offloads more string munging into go-containerregistry which makes it their (er, my) responsibility to test 😄

pkg/apis/policy/v1beta1/clusterimagepolicy_validation.go
```diff
@@ -219,6 +224,20 @@ func (source *Source) Validate(ctx context.Context) *apis.FieldError {
 	var errs *apis.FieldError
 	if source.OCI == "" {
 		errs = errs.Also(apis.ErrMissingField("oci"))
+	} else {
+		// We want to validate both registry uris only or registry with valid repository names
```
vaikas, line 228: I'm fine with this small change but I think I'd still prefer to pull this into a function since it's still cut&paste in both places :) But I'm not going to block it. I'm just thinking that once we roll to a new version later on, we'll have to do this then too, so might as well do it now. But up to you :)

vaikas (approved):

pkg/apis/policy/v1alpha1/clusterimagepolicy_validation.go
hectorj2f: It's an option. But I'd avoid appending any fake value to what we get from the users. 
We still offload all the validations while using generic calls to count the parts.

pkg/apis/policy/v1beta1/clusterimagepolicy_validation.go
hectorj2f: I'll take care of that in a different PR so I can merge these changes now.

hectorj2f: I've addressed here: https://github.com/sigstore/policy-controller/pull/341

PR #340, opened 2022-10-22
Fix: Always use `kubeclient.Get()` for fetching k8s client.

mattmoor:
:bug: I was debugging an issue in a downstream call to `ValidatePolicy` where I noticed that we already expect `SourceSignaturePullSecretsOpts` to fetch the kubernetes client off of `ctx` via `kubeclient.Get(ctx)`.

Given that this is a fairly uncommon code path to exercise, I wanted to make the way we access the client more consistent so that downstream callers discover this issue more quickly and can fix the problem before it hits in the corner case they likely have not tested.

/kind bug

#### Release Note

Fixes a crash during the validation webhook when image pull secrets are used.

### DIFF
--- a/cmd/webhook/main.go
+++ b/cmd/webhook/main.go
@@ -29,6 +29,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	duckv1 "knative.dev/pkg/apis/duck/v1"
+	kubeclient "knative.dev/pkg/client/injection/kube/client"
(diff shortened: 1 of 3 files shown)


mattmoor:
I've updated this to include a fix for an issue where if the `SourceSignaturePullSecretsOpts` is used, then policy validation will panic because the context is not injected with a kubeclient.

mattmoor:
looks like a flake

cpanato (approved):
Nice

hectorj2f (approved):
Thanks @mattmoor

PR #346, opened 2022-10-25
Fix: switch from `all` to `ALL` when dropping capabilities.

mattmoor:
:bug: TIL that `all` isn't recognized by K8s upstream, and they hardcode an `ALL` constant

ref: https://github.com/knative-sandbox/eventing-rabbitmq/pull/954#pullrequestreview-1153348371

/kind bug

#### Release Note

Change the way we drop capabilities from `all` to `ALL`, which is the constant K8s expects.

#### Documentation

cc @hectorj2f @vaikas

### DIFF
--- a/config/policy-webhook.yaml
+++ b/config/policy-webhook.yaml
@@ -72,7 +72,7 @@ spec:
           runAsNonRoot: true
           capabilities:
             drop:
-            - all
+            - ALL
 
         volumeMounts:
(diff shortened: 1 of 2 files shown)


k4leung4 (approved):

hectorj2f (approved):

hectorj2f:
We'll need to change that in the helm chart too.

PR #348, opened 2022-10-26
Add policy name to the cache

hectorj2f:

#### Summary

We're just storing the CIP uid which doesn't give a descriptive information of which policy was evaluated. I suggest storing both uid and name, so we can give more detailed information of the results after evaluating a policy.

#### Release Note

Add policy name to the cache fields.


cpanato (approved):

hectorj2f:
pinging @tcnghia here

tcnghia (approved):

mattmoor (approved):
This is breaking at a Go library level, but new enough that I think it's worth breaking early/now.  thanks!
-M

PR #349, opened 2022-10-26
Set no matching policy flag in the cache entry

hectorj2f:

#### Summary

We decided to cache the results when an image does not apply to any policy. To do so, we add a flag to the Set(...) function. 

#### Release Note

Add a flag `NoMatchingPolicy` to the cache entries.

### DIFF
--- a/pkg/webhook/cache.go
+++ b/pkg/webhook/cache.go
@@ -46,10 +46,11 @@ func ToContext(ctx context.Context, cache ResultCache) context.Context {
 
 type ResultCache interface {
 	// Set caches a PolicyResult for a given CIP evaluated for a given image at
-	// a particular point in time. image, uid & resourceVersion will give a
+	// a particular point in time. image, uid, name & resourceVersion will give a
 	// unique point in time, so we can make sure we're not caching things that
-	// are out of date.
-	Set(ctx context.Context, image, name, uid, resourceVersion string, cacheResult *CacheResult)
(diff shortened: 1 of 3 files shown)


tcnghia (approved):

pkg/webhook/cache.go
```diff
@@ -46,10 +46,11 @@ func ToContext(ctx context.Context, cache ResultCache) context.Context {
...
 type ResultCache interface {
 	// Set caches a PolicyResult for a given CIP evaluated for a given image at
-	// a particular point in time. image, uid & resourceVersion will give a
+	// a particular point in time. image, uid, name & resourceVersion will give a
 	// unique point in time, so we can make sure we're not caching things that
-	// are out of date.
-	Set(ctx context.Context, image, name, uid, resourceVersion string, cacheResult *CacheResult)
+	// are out of date. We need to cache any image that does not match any policy,
+	// thereby this function has a parameter noMatchingPolicy to specify it.
+	Set(ctx context.Context, image, name, uid, resourceVersion string, noMatchingPolicy bool, cacheResult *CacheResult)
```
tcnghia, line 53: Since the `CacheResult.PolicyResult` field is `nil/empty` in these cases, we should already know that this is due to no match and don't need to change the signature?

hectorj2f:
Closing this PR! We are exploring other alternatives.

PR #350, opened 2022-10-27
Feature: Incorporate an identifier for signatures and attestations.

mattmoor:
:gift: This incorporates an identifier for signatures and attestations into a Policy Result, which should be suitable for identifying a signature or attestation across policy invocations (possibly across multiple policies).

/kind feature

#### Release Note

`PolicyResult` now includes an `Id` in `Policy{Signature,Attestation}` suitable for identifying the signature or attestation across contexts.

### DIFF
--- a/pkg/webhook/validator_result.go
+++ b/pkg/webhook/validator_result.go
@@ -55,6 +55,10 @@ type AuthorityMatch struct {
 // signature could be a signature on the Image (.sig) or on an Attestation
 // (.att).
 type PolicySignature struct {
+	// A unique identifier describing this signature.
+	// This is typically the hash of this signature's OCI layer for images.
+	ID string `json:"id,omitempty"`
+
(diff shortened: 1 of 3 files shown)


hectorj2f (approved):

pkg/webhook/validator_result.go
```diff
@@ -55,6 +55,10 @@ type AuthorityMatch struct {
 // signature could be a signature on the Image (.sig) or on an Attestation
 // (.att).
 type PolicySignature struct {
+	// A unique identifier describing this signature.
+	// This is typically the hash of this signature's OCI layer for images.
+	ID string `json:"id,omitempty"`
```
hectorj2f, line 60: What about adding an ID for the PolicyAttestation too ? It could be the same ID of the PolicySignature it consists of.

PR #359, opened 2022-10-31
fix: error message

hectorj2f:

#### Summary

I found some errors in my logs related to this error message: `could not decode body:%!(EXTRA *errors.errorString=EOF)`


pkg/webhook/validator.go
```diff
@@ -903,7 +903,7 @@ func getNamespace(ctx context.Context, namespace string) string {
 		if r != nil && r.Body != nil {
 			var review admissionv1.AdmissionReview
 			if err := json.NewDecoder(r.Body).Decode(&review); err != nil {
-				logging.FromContext(ctx).Errorf("could not decode body:", err)
+				logging.FromContext(ctx).Errorf("could not decode body: %w", err)
```
vaikas, line 906: Does the wrap directive work with the Errorf? I'd expect %v in the logging statement. But if you're certain it works, alrighty :)

vaikas (dismissed):

pkg/webhook/validator.go
hectorj2f: Good catch! I was expecting zap to handle it, but it doesn't all well.

cpanato (approved):
great catch!

PR #362, opened 2022-11-02
fix: allow spec.authorities field to not be specified

### LINKED ISSUE #361: The authorities field cannot be left out despite setting fail-on-empty-authorities
**Description**

When `fail-on-empty-authorities: "false"` (see #275 and #292 for details), the validation still fails with:

```
The ClusterImagePolicy "..." is invalid: spec.authorities: Invalid value: "null": spec.authorities in body must be of type array: "null"
```

This is because the field is expected.

The fix should be as simple as adding `omitempty` to the field definition, marking it as optional.

**Version**

Verified on latest `main` and release 0.4.0.

wojciechka:
Closes #361 

#### Summary

Add `omitempty` to `authorities` to allow the `spec.authorities` field to not be specified.

This field being present and not nil is already conditionally validated (with the validation happening by default), so this should be safe to add.

To test the change:

1. Ensure `policy-controller` configuration includes `fail-on-empty-authorities: "false"`
2. Apply the following YAML:
```
kind: ClusterImagePolicy
metadata:
  name: empty-authorities
spec:
  images:
  - glob: 'index.docker.io/*'
EOF
```

This should work, fails on latest main, but works with the code change from this PR.

#### Release Note

Allow `spec.authorities` field to not be specified

#### Documentation

N/A

### DIFF
--- a/pkg/apis/policy/v1alpha1/clusterimagepolicy_types.go
+++ b/pkg/apis/policy/v1alpha1/clusterimagepolicy_types.go
@@ -55,6 +55,7 @@ type ClusterImagePolicySpec struct {
 	// Images defines the patterns of image names that should be subject to this policy.
 	Images []ImagePattern `json:"images"`
 	// Authorities defines the rules for discovering and validating signatures.
+	// +optional
 	Authorities []Authority `json:"authorities,omitempty"`
 	// Policy is an optional policy that can be applied against all the
 	// successfully validated Authorities. If no authorities pass, this does
--- a/pkg/apis/policy/v1alpha1/clusterimagepolicy_types.go
+++ b/pkg/apis/policy/v1alpha1/clusterimagepolicy_types.go
@@ -55,7 +55,7 @@ type ClusterImagePolicySpec struct {
 	// Images defines the patterns of image names that should be subject to this policy.
 	Images []ImagePattern `json:"images"`
 	// Authorities defines the rules for discovering and validating signatures.
-	Authorities []Authority `json:"authorities"`
+	Authorities []Authority `json:"authorities,omitempty"`
 	// Policy is an optional policy that can be applied against all the
 	// successfully validated Authorities. If no authorities pass, this does
 	// not even get evaluated, as the Policy is considered failed.
--- a/pkg/apis/policy/v1beta1/clusterimagepolicy_types.go
+++ b/pkg/apis/policy/v1beta1/clusterimagepolicy_types.go
@@ -55,6 +55,7 @@ type ClusterImagePolicySpec struct {
 	// Images defines the patterns of image names that should be subject to this policy.
 	Images []ImagePattern `json:"images"`
 	// Authorities defines the rules for discovering and validating signatures.
+	// +optional
 	Authorities []Authority `json:"authorities,omitempty"`
 	// Policy is an optional policy that can be applied against all the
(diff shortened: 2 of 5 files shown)


hectorj2f:
You shared an error that is coming from the CRD definition where the type needs to be an array.

hectorj2f:
@wojciechka you need to signoff your commits

hectorj2f (approved):

pkg/apis/policy/v1alpha1/clusterimagepolicy_types.go
```diff
@@ -55,7 +55,7 @@ type ClusterImagePolicySpec struct {
 	// Images defines the patterns of image names that should be subject to this policy.
 	Images []ImagePattern `json:"images"`
 	// Authorities defines the rules for discovering and validating signatures.
-	Authorities []Authority `json:"authorities"`
+	Authorities []Authority `json:"authorities,omitempty"`
```
vaikas, line 58: nit: Could we add the
```
// +optional
```

before L58

pkg/apis/policy/v1beta1/clusterimagepolicy_types.go
```diff
@@ -55,7 +55,7 @@ type ClusterImagePolicySpec struct {
 	// Images defines the patterns of image names that should be subject to this policy.
 	Images []ImagePattern `json:"images"`
 	// Authorities defines the rules for discovering and validating signatures.
-	Authorities []Authority `json:"authorities"`
+	Authorities []Authority `json:"authorities,omitempty"`
```
vaikas, line 58: Ditto.

vaikas (commented):
Thanks for doing this! Could we also add two example test files here (one for v1alpha1 and one for v1beta1)?
https://github.com/sigstore/policy-controller/tree/main/test/testdata/policy-controller/valid

pkg/apis/policy/v1alpha1/clusterimagepolicy_types.go
wojciechka: Done in both files

wojciechka:
> Thanks for doing this! Could we also add two example test files here (one for v1alpha1 and one for v1beta1)?
> https://github.com/sigstore/policy-controller/tree/main/test/testdata/policy-controller/valid

This is controlled by the `fail-on-empty-authorities` configuration option and not setting `spec.authorities` is not valid by default - you need to set `fail-on-empty-authorities` to `false` and only then it becomes possible.

There's even [test/testdata/policy-controller/invalid/v1beta1-keylessref-with-empty-authorities.yaml](https://github.com/sigstore/policy-controller/blob/main/test/testdata/policy-controller/invalid/v1beta1-keylessref-with-empty-authorities.yaml) that tests this is not valid `ClusterImagePolicy`.

hectorj2f:
@wojciechka I think we want to test the scenario we had when setting `fail-on-empty-authorities` to false where we weren't able to update a CIP. With these changes, we can now update an initial CIP without authorities.

wojciechka:
> I will suggest adding here the following test https://github.com/sigstore/policy-controller/blob/main/test/e2e_test_policy_crd.sh#L114

Thank you. Added the logic at the end of shell script and added test for both existing and non-existing `spec.authorities` in the YAML file.

hectorj2f (approved):
lgtm

PR #364, opened 2022-11-04
feat: optionally allow updating image to digest only at Pod level

### LINKED ISSUE #285: Allow configuring whether image is replaced with SHA digest
The `policy-controller` mutating webhook is patching images in `PodSpec` to replace tags with digests - for example `docker.io/library/telegraf:1.22` in Deployment is expanded to `index.docker.io/library/telegraf@sha256:05b59d689b7812cfc31a3a2eb70fe18d3adea74a1d572fbf1a91294e091359e7`.

I understand the rationale behind it, but it would be great to allow some configuration to prevent this behavior either at workload, namespace, or global level.

Applying manifests via CD (or probably other reconciliation tools as well) will cause the tool to go into indefinite loop of trying to update the image. This happened in an environment with ArgoCD `v2.2.11` - where ArgoCD was reporting a difference between expected and live manifest and trying to reconcile it, only to be updated by the webhook.

This is fine when having control over the manifests - it forces pinning to specific SHA digest.

However, we sometimes use CD for deploying workloads we do not fully control or want to control - such as Helm charts or other third-party manifests where the image may be specified as tag, and this will cause CD to go into an infinite loop/conflict with `policy-controller`.

It would be great if this would be configurable, preferably not globally as the behavior may be forcing good practices in some cases.

Confirmed on version 0.3.0

wojciechka:
#### Summary

Allow configuring policy-controller to only update images at PodSpec level only. This avoids certain CDs considering image updates at Deployment, StatefulSet or other higher-level object levels as drifts.

Closes #285

#### Release Note
<!--
Add a release note for each of the following conditions:

* Allow configuring policy-controller to only update images at PodSpec level only

#### Documentation

N/A

### DIFF
--- a/pkg/webhook/validator.go
+++ b/pkg/webhook/validator.go
@@ -88,7 +88,7 @@ func (v *Validator) ValidatePodScalable(ctx context.Context, ps *policyduckv1bet
 		ImagePullSecrets:   imagePullSecrets,
 	}
 
-	return v.validatePodSpec(ctx, ns, ps.Kind, ps.APIVersion, ps.ObjectMeta.Labels, &ps.Spec.Template.Spec, opt).ViaField("spec.template.spec")
+	return v.validatePodSpec(ctx, ns, ps.Kind, ps.APIVersion, ps.ObjectMeta.Labels, &ps.Spec.Template.Spec, opt, false).ViaField("spec.template.spec")
 }
 
 // ValidatePodSpecable implements duckv1.PodSpecValidator
@@ -108,7 +108,7 @@ func (v *Validator) ValidatePodSpecable(ctx context.Context, wp *duckv1.WithPod)
 		ServiceAccountName: wp.Spec.Template.Spec.ServiceAccountName,
 		ImagePullSecrets:   imagePullSecrets,
 	}
-	return v.validatePodSpec(ctx, ns, wp.Kind, wp.APIVersion, wp.ObjectMeta.Labels, &wp.Spec.Template.Spec, opt).ViaField("spec.template.spec")
+	return v.validatePodSpec(ctx, ns, wp.Kind, wp.APIVersion, wp.ObjectMeta.Labels, &wp.Spec.Template.Spec, opt, false).ViaField("spec.template.spec")
 }
 
 // ValidatePod implements duckv1.PodValidator
@@ -128,7 +128,7 @@ func (v *Validator) ValidatePod(ctx context.Context, p *duckv1.Pod) *apis.FieldE
 		ServiceAccountName: p.Spec.ServiceAccountName,
 		ImagePullSecrets:   imagePullSecrets,
 	}
-	return v.validatePodSpec(ctx, ns, p.Kind, p.APIVersion, p.ObjectMeta.Labels, &p.Spec, opt).ViaField("spec")
+	return v.validatePodSpec(ctx, ns, p.Kind, p.APIVersion, p.ObjectMeta.Labels, &p.Spec, opt, true).ViaField("spec")
 }
 
 // ValidateCronJob implements duckv1.CronJobValidator
@@ -149,10 +149,13 @@ func (v *Validator) ValidateCronJob(ctx context.Context, c *duckv1.CronJob) *api
 		ImagePullSecrets:   imagePullSecrets,
 	}
 
-	return v.validatePodSpec(ctx, ns, c.Kind, c.APIVersion, c.ObjectMeta.Labels, &c.Spec.JobTemplate.Spec.Template.Spec, opt).ViaField("spec.jobTemplate.spec.template.spec")
+	return v.validatePodSpec(ctx, ns, c.Kind, c.APIVersion, c.ObjectMeta.Labels, &c.Spec.JobTemplate.Spec.Template.Spec, opt, false).ViaField("spec.jobTemplate.spec.template.spec")
 }
 
-func (v *Validator) validatePodSpec(ctx context.Context, namespace, kind, apiVersion string, labels map[string]string, ps *corev1.PodSpec, opt k8schain.Options) (errs *apis.FieldError) {
+func (v *Validator) validatePodSpec(ctx context.Context, namespace, kind, apiVersion string, labels map[string]string, ps *corev1.PodSpec, opt k8schain.Options, isPod bool) (errs *apis.FieldError) {
+	pcConfig := policycontrollerconfig.FromContextOrDefaults(ctx)
+	logging.FromContext(ctx).Warnf("TEST00 - validate - %v", pcConfig.ReplaceDigestInPodOnly)
+
 	kc, err := k8schain.New(ctx, kubeclient.Get(ctx), opt)
 	if err != nil {
 		logging.FromContext(ctx).Warnf("Unable to build k8schain: %v", err)
@@ -173,12 +176,15 @@ func (v *Validator) validatePodSpec(ctx context.Context, namespace, kind, apiVer
 			go func() {
 				defer wg.Done()
 
-				// Require digests, otherwise the validation is meaningless
-				// since the tag can move.
-				_, fe := refOrFieldError(c.Image, field, i)
-				if fe != nil {
-					results <- containerCheckResult{index: i, containerCheckResult: fe}
-					return
+				// validate either if images should always be digests or this is PodSpec
+				if isPod || !pcConfig.ReplaceDigestInPodOnly {
+					// Require digests, otherwise the validation is meaningless
+					// since the tag can move.
+					_, fe := refOrFieldError(c.Image, field, i)
+					if fe != nil {
+						results <- containerCheckResult{index: i, containerCheckResult: fe}
+						return
+					}
 				}
 
 				containerErrors := v.validateContainerImage(ctx, c.Image, namespace, field, i, kind, apiVersion, labels, ociremote.WithRemoteOptions(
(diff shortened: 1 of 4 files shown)


wojciechka:
Created as draft for now to review if the approach and outcome makes sense before implementing additional tests for this.

With this change and `"replace-digest-in-pod-only": "true"`, deploying:

```
apiVersion: apps/v1
kind: Deployment
metadata:
  ...
spec:
  template:
    spec:
      containers:
        - image: docker.io/library/telegraf:1.22
...
```

Causes Deployment to continue having `image: docker.io/library/telegraf:1.22`, but all Pods have their image updated to `image: index.docker.io/library/telegraf@sha256:3dc951b21437d921f4088baa8e04022f0314eeeabae80619e92bc11dd1f79bcb`.

This will be slower, so it is opt-in, but I do not think this decreases security - the image is still updated and validated at Pod level, it will simply be slower for Deployments with multiple replicas as the image will be resolved for each new Pod.

hectorj2f (commented):
I think you could achieve the same effect. If you create CIPs that only match Pods and have this configuration flag enabled, so the digest would be converted for the pods.

pkg/webhook/validator.go
```diff
@@ -149,10 +149,13 @@ func (v *Validator) ValidateCronJob(ctx context.Context, c *duckv1.CronJob) *api
 		ImagePullSecrets:   imagePullSecrets,
 	}
 
-	return v.validatePodSpec(ctx, ns, c.Kind, c.APIVersion, c.ObjectMeta.Labels, &c.Spec.JobTemplate.Spec.Template.Spec, opt).ViaField("spec.jobTemplate.spec.template.spec")
+	return v.validatePodSpec(ctx, ns, c.Kind, c.APIVersion, c.ObjectMeta.Labels, &c.Spec.JobTemplate.Spec.Template.Spec, opt, false).ViaField("spec.jobTemplate.spec.template.spec")
 }
 
-func (v *Validator) validatePodSpec(ctx context.Context, namespace, kind, apiVersion string, labels map[string]string, ps *corev1.PodSpec, opt k8schain.Options) (errs *apis.FieldError) {
+func (v *Validator) validatePodSpec(ctx context.Context, namespace, kind, apiVersion string, labels map[string]string, ps *corev1.PodSpec, opt k8schain.Options, isPod bool) (errs *apis.FieldError) {
```
hectorj2f, line 155: This is serving a very specific use case, I thought you wanted to disable the mutation of the digest for all the resources.

wojciechka: I was thinking about it and realized that this is not generic enough in its form.

What I think this should be is making digest behavior configurable to specify when the image is replaced with digest - in all objects (Deployment, StatefulSet, CronJob, Job, ... as well as Pod), pod or never.

So from configuration perspective it could be the following instead of current setting in this PR:

```
update-image-to-digests: always
update-image-to-digests: pod
update-image-to-digests: never
```
(defaulting to `always`)

This way it's possible to configure whether top-level object like Deployment gets mutated (default, fastest, secure), only the Pods get mutated (which is slower, but should still be as secure), no objects get mutated (which is less secure).

I think for the setup that we have we would like the Pod-level mutation to happen, but others may want something else, so having 3 options should be most configurable.

As for passing this line specifically and `isPod`, I'm simply passing if this was called for Pod or another type of object so the code can determine its behavior.

Would that make more sense?

tysonkamp: I'm not the k8s expert but "never" sure sounds risky.

wojciechka: I have changed this approach and pushed the change. The only difference from my comment above is I called th setting `update-image-to-digest`. It acceps settings `always`, `pod` and `never`.

This way it defaults to current behavior, but allows configuring whether to patch Pods only or not patch at all.

I think this is a reasonable feature and not just specific to our use case. In many cases it's fine to patch derived objects such as Pod, but it is more complex to have objects managed by GitOps patched by other controllers / operators.

I find the pattern of patching Pods instead of higher-level objects like Deployment/StatefulSet/CronJob/Job quite common. In this case it seems like a reasonable compromise to me - patching Pods prevents image tag updates between validation and pod starting from being a security issue.

vaikas: I'm really curious why you would want to patch Pod A of Deployment X to be different than say Pod B. Just trying to understand why this is common. Naively this seems to me as an antipattern, so just trying to understand :)

hectorj2f: @imjasonh Do you have any ideas how to handle this use case ? We believe you have more thoughts on why we should  avoid using tags, and instead stick to digests.

wojciechka: I've been looking more into this, trying to understand how this is a problem for me, how other tools to it and think of ways forward. It helped me clear my mind.

The projects I work on use GitOps and ArgoCD to deploy everything - there's a repo that has everything included, dependencies vendored in and rendered as YAML files. ArgoCD then deploys those files and makes sure that all Kubernetes objects in manifests that are in the Git repo (such as Deployments, StatefulSets etc). Some of our deployments are upstream projects we use, either via helm charts or in other ways.

When `policy-controller` and its mutating webhooks, whenever ArgoCD updates or creates objects like Deployments, it patches the image, and this is the live manifest of a Deployment. ArgoCD sees this as a difference between desired and live manifest - it expects `some/image:v1.23` while Kubernetes returns `docker.io/some/image@sha256:...`. It causes ArgoCD to be consistently trying to apply the difference and never considers the application to be synchronized and up to date.

This only happens for any objects that do not use digests and in an ideal world I would simply only use image digests. We have over 100 apps we are including in our environments, many of which are just upstream charts or manifests. I did a quick grep and am seeing several hundred different images not referenced via digests across all apps and environments.

In many cases there's no simple and standard way to set images to digests - so I'd rather avoid having to find ways to patch the YAML manifests to patch any image tags to image digests to use `policy-controller`.

Unfortunately this is often the case when trying to apply a new tool to existing environments.

As for the why - I think the problem is that ArgoCD is expecting manifests to be applied without mutations (at least the level of objects ArgoCD manages - i.e. it ignores any changes/mutations at the level of ReplicaSets/Pods), while `policy-controller` today applies mutations to all objects in namespaces it controls. I am not sure why I seem to be the first one reporting it - I would expect this to be a more common problem with GitOps approach. ArgoCD (and I suppose other CD tools) work best when mutations are stored in new fields, not replacing existing fields.

I see ArgoCD allows ignoring fields managed by other controllers, but I don't think changes from mutating webhooks are tracked as managed fields or if there is an easy way to do so. That could be too complex to implement, though and I am not sure if it would work in this specific case - docs are here: https://argo-cd.readthedocs.io/en/stable/user-guide/diffing/#application-level-configuration

So at that point I was curious about other options. I don't think we'd be willing to look into alternatives for ArgoCD just because of this issue. But I checked [customizing how ArgoCD is diffing objects](https://argo-cd.readthedocs.io/en/stable/user-guide/diffing/) in ArgoCD and if it could help, but this causes another problem - if the image changes are ignored and are the only change in a Deployment, ArgoCD will just not apply the new image.

It seems that [connaisseur](https://github.com/sse-secure-systems/connaisseur/) started by only patching and validating Pods. There's a feature called [Automatic Child Approval](https://github.com/sse-secure-systems/connaisseur/blob/5574d34f86dfba8622d173f5cc159bd64baa6038/docs/features/automatic_child_approval.md) that does something similar to policy-controller's current behavior - replacing image with digest at the parent level.

Ideally, I think the solution should be done at the node and/or Kubernetes level - ideally Kubernetes in managed clouds should provide a way to validate images (either using webhooks or similar mechanism) so `policy-controller` could simply reject images after they're pulled on the node, but before pod is started. I remember someone at SigstoreCon mentioned that this is how they are doing image validation, but it works because they are not using managed Kubernetes clusters. Unfortunately, this is not something that would be available in a reasonable timeframe, even if work on it would begin now.

Another solution would be for `policy-controller` to resolve images at Deployment level but store them as PodSpec's annotation or any other location and then use that at the Pod level. This annotation could also have a signature based on webhook's private/public key pair. I think this could be too complex for the problem I am trying to find a solution to.

So, this makes me go back to my solution. I think it's reasonable, it allows 3 modes of operation - current mode, which always updates all objects, pod only updates and no updates. The first one makes sense for new workloads. The `pod` setting makes it behave more like [connaisseur](https://github.com/sse-secure-systems/connaisseur/) that it only patches Pods and not parent objects, but it also validates parent objects, which is good as it makes applying a Deployment with invalid images fail right away. I am not that passionate about `never` option, but I added it for completeness.

Regarding pod level patching - the main problem there is additional round trips to OCI registries. Resolving tag to digest at each `Pod` level is going to add more load, but I think that can be improved by adding caching.

I already created #278 as `policy-controller` is resolving tag to digest and then checking digest's signatures for every key in key-based validation. I think there's some cache scaffolding in the code already. Perhaps caching retrieving signatures for a digest and resolving tag to digest could be cached and then the overhead of patching each pod instead of patching the parent would not be as high.

As for pods potentially having different patches applied - in most cases the tag-based images are pointing to specific releases, not rolling tags. So if the images do not change, each pod would have the same patch applied to it. This is also how creating deployments with tag-based images works today - if I deploy `ubuntu:latest`, a pod created one year ago and today would potentially have different image.

wojciechka:
There is no consensus on implementation details on this for now, so I'll close the PR.

For my specific case, we'll try to opt-in namespaces for policy-controller at a slower pace, addressing any images not referenced via SHA digest. This may be the correct approach in the long term.

Issue #382, opened 2022-11-12
Drop service account lookups when `signaturePullSecrets` are specified


mattmoor:
**Description**

Currently in `k8schain` it is ~impossible to avoid looking up pull secrets attached to a service account, however, I am plumbing through that option in: https://github.com/google/go-containerregistry/pull/1490

I believe that when `signaturePullSecrets` are explicitly specified that we should pass this new constant for `ServiceAccountName` here:
https://github.com/sigstore/policy-controller/blob/87267f39c925c0922c89e042980df43ef0bbe91e/pkg/webhook/clusterimagepolicy/clusterimagepolicy_types.go#L203-L206

... this request creates unnecessary load on the API server.

hectorj2f:
@mattmoor 100% agree, we should avoid doing that call!

Issue #360, opened 2022-11-01
Multiple `source:` likely won't work as intended


mattmoor:
**Description**

I haven't explicitly tried this, but I'm fairly confident that this is a problem looking at the code.  Currently `source:` takes a list, but when we're unmarshalling things from our "compiled" form we call `WithTargetRepository` and I think the last one wins:

https://github.com/sigstore/policy-controller/blob/970ac4a04700727fe984376d878fbec1d0bcdcb5/pkg/webhook/clusterimagepolicy/clusterimagepolicy_types.go#L182

There's a related problem where we just sort of YOLO our way through the credentials and last-one wins here:

https://github.com/sigstore/policy-controller/blob/970ac4a04700727fe984376d878fbec1d0bcdcb5/pkg/webhook/clusterimagepolicy/clusterimagepolicy_types.go#L196-L197

I thought maybe we might be restricting the `source:` to a singleton for now, but that doesn't appear to be the case (and I was just talking to @dlorenc about how useful this being a list could be!):
https://github.com/sigstore/policy-controller/blob/970ac4a04700727fe984376d878fbec1d0bcdcb5/pkg/apis/policy/v1beta1/clusterimagepolicy_validation.go#L142-L144

cc @hectorj2f @vaikas 

**Version**

HEAD, v0.4.1

vaikas:
I'm going to restrict the validation to be a singleton entry in the list until we actually support it properly. It's a bigger change and I reckon that having it in a bad state is no good.
@mattmoor @hectorj2f

PR #388, opened 2022-11-16
Drop service account lookups when signaturePullSecrets are specified

hectorj2f:
#### Summary

fixes: https://github.com/sigstore/policy-controller/issues/382 

#### Release Note

Drop service account lookups when signaturePullSecrets are specified.


pkg/webhook/validator_test.go
```diff
@@ -573,19 +573,7 @@ func TestValidatePodSpec(t *testing.T) {
 				},
 			},
 		),
-		want: func() *apis.FieldError {
```
hectorj2f, line 576: If a secret (imagePullSecrets) or service account doesn't exist, the package go-containerregistry ignores these errors https://github.com/google/go-containerregistry/issues/1471.

vaikas:
It looks like for the sources, we have zero coverage on our e2e tests, could we add at least one failing and one passing test there since it (from a quick look) our unit tests now all turn into ignores with the sources imagepull secrets.
FWIF, I think I'd prefer to create a new e2e_test_cluster_image_policy_with_sources or something like that instead of adding to existing ones.

vaikas:
Also would be handy to start adding tests for sources for things like this as well:
https://github.com/sigstore/policy-controller/issues/360

hectorj2f:
@vaikas Yes, I like that idea.

pkg/webhook/validator_test.go
```diff
@@ -2239,19 +2219,7 @@ func TestValidatePodSpecNonDefaultNamespace(t *testing.T) {
...
```
vaikas, line 2242: Just wondering if there's any point in keeping these tests around?

test/e2e_test_cluster_image_policy_with_source.sh
```diff
@@ -0,0 +1,132 @@
...
+
+echo '::group:: Change source.oci in policy to a wrong repository without signatures'
+kubectl patch cip image-policy-keyless-source --type "json" \
+-p '[{"op":"replace", "path":"/spec/authorities/0/source/0/oci", "value":"ghcr.io/sigstore/cosign/cosign"}]'
+# allow for propagation
+sleep 5
+echo '::endgroup::'
+
+echo '::group:: test job success as source.oci points to the right repository'
+# We signed this above, this should work
```
vaikas, line 119: nit, this comment is wrong? This should fail because the it's trying to fetch signatures that don't exist.

test/e2e_test_cluster_image_policy_with_source.sh
```diff
@@ -0,0 +1,132 @@
...
+if kubectl create -n demo-keyless-source job demo2 --image=${demoimage} ; then
+  echo Failed to block Job creation with wrong OCI source!
```
vaikas, line 121: I think it would be nice to check the error message that we get to make sure the error is what we expect and not some other failure.
You can use something like this:
https://github.com/sigstore/policy-controller/blob/main/test/e2e_test_cluster_image_policy_with_attestations.sh#L66

and then call it like this:
https://github.com/sigstore/policy-controller/blob/main/test/e2e_test_cluster_image_policy_with_attestations.sh#L203

vaikas (dismissed):
Thanks for adding the tests!! Couple of nits but I'm fine as is.

pkg/webhook/validator_test.go
hectorj2f: I think it is fine to keep them so we can exercise these all these fields in case there are errors when using go-containerregistry in the future.

test/e2e_test_cluster_image_policy_with_source.sh
hectorj2f: Done, thx for the suggestions ❤️ !

vaikas (approved):

Issue #385, opened 2022-11-15
Allow policies to access image config(s) and resource spec


mattmoor:
**Description**

_This is capturing a convo @dlorenc and I were having for more discussion._

Two useful pieces of information for policy evaluation, which are readily available to the webhook, but the policy language does not currently have access to are:
1. The image's ["config file"](https://github.com/google/go-containerregistry/blob/781782aa2757f8eb6d3b7b7a32b25a799b64bc60/pkg/v1/config.go#L23)
2. The resource's `spec`

To not bury the lede...  Personally, I think the most straightforward option would be to simply include these two pieces of information as part of `PolicyResult`: https://github.com/sigstore/policy-controller/blob/ba6cdef1d82974440ea2eb23029fe6150bceb265/pkg/webhook/validator_result.go#L18-L36

I think we could add a `ResourceSpec interface{}` (with the `.spec` from the resource) and `Config map[string]v1.ConfigFile` (the string key for Config is to support [multi-arch images](https://github.com/google/go-containerregistry/blob/781782aa2757f8eb6d3b7b7a32b25a799b64bc60/pkg/v1/platform.go#L23-L31)).

This would let us access this context as part of the top-level `Policy` block here:
https://github.com/sigstore/policy-controller/blob/ba6cdef1d82974440ea2eb23029fe6150bceb265/pkg/apis/policy/v1beta1/clusterimagepolicy_types.go#L60-L64

### Other Considerations (incl. alternatives)

Including the resource spec as part of the policy result felt like a natural extension of the relatively new `Match` capabilities, but allowing users to BYO free form validation over the resource specification: https://github.com/sigstore/policy-controller/blob/ba6cdef1d82974440ea2eb23029fe6150bceb265/pkg/apis/policy/v1beta1/clusterimagepolicy_types.go#L71-L73

We could make providing this information opt-in via fields (e.g. `includeConfig: true`, `includeSpec: true`), I don't love the names, so if you have better names, feel free to offer up alternatives.  I'd also be fine with simply always including this information.

One alternative for the config file would be to have users project it as an attestation, but this felt awkward and frankly redundant since the config file is already associated with the container and part of the signed merkle tree.

Another alternative for the config file was to introduce a special case `predicateType` that accessed the image's embedded config  instead of requiring an explicit attestation, but this feels a bit to magical IMO.

hectorj2f:
@vaikas Should we close this issue now ?

mattmoor:
[...]

PR #389, opened 2022-11-17
Add FetchConfigFile to Policy that allows you to fetch and evaluate policy against container image configfile.

vaikas:

#### Summary
Addresses the configfile part of #385

Add flag to spec.policy `fetchConfigFile` that when set to true will fetch and make the ConfigFile available for CIP level
policy evalutions.

#### Release Note
 * Add FetchConfigFile into cip.Spec.Policy that allows one to fetch OCI config files (https://github.com/opencontainers/image-spec/blob/main/config.md) that are then available for evaluation in the cip.Spec.Policy

### DIFF
--- a/pkg/apis/policy/v1alpha1/clusterimagepolicy_types.go
+++ b/pkg/apis/policy/v1alpha1/clusterimagepolicy_types.go
@@ -203,7 +203,8 @@ type MatchResource struct {
 	ResourceSelector *metav1.LabelSelector `json:"selector,omitempty"`
 }
 
-// Policy specifies a policy to use for Attestation validation.
+// Policy specifies a policy to use for Attestation or the CIP validation (iff
+// at least one authority matches).
 // Exactly one of Data, URL, or ConfigMapReference must be specified.
 type Policy struct {
 	// Which kind of policy this is, currently only rego or cue are supported.
@@ -218,6 +219,13 @@ type Policy struct {
 	// ConfigMapRef defines the reference to a configMap with the policy definition.
 	// +optional
 	ConfigMapRef *ConfigMapReference `json:"configMapRef,omitempty"`
+	// FetchConfigFile controls whether ConfigFile will be fetched and made
+	// available for CIP level policy evaluation. Note that this only gets
+	// evaluated (and hence fetched) iff at least one authority matches.
+	// The ConfigFile will then be available in this format:
+	// https://github.com/opencontainers/image-spec/blob/main/config.md
+	// +optional
+	FetchConfigFile *bool `json:"fetchConfigFile,omitempty"`
 }
 
 // ConfigMapReference is cut&paste from SecretReference, but for the life of me
--- a/pkg/webhook/validator.go
+++ b/pkg/webhook/validator.go
@@ -22,7 +22,11 @@ import (
 	"fmt"
 	"sync"
 
+	"github.com/google/go-containerregistry/pkg/authn"
 	"github.com/google/go-containerregistry/pkg/authn/k8schain"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/types"
+
 	"github.com/google/go-containerregistry/pkg/name"
 	"github.com/google/go-containerregistry/pkg/v1/remote"
 	"github.com/sigstore/cosign/pkg/cosign"
@@ -181,7 +185,7 @@ func (v *Validator) validatePodSpec(ctx context.Context, namespace, kind, apiVer
 					return
 				}
 
-				containerErrors := v.validateContainerImage(ctx, c.Image, namespace, field, i, kind, apiVersion, labels, ociremote.WithRemoteOptions(
+				containerErrors := v.validateContainerImage(ctx, c.Image, namespace, field, i, kind, apiVersion, labels, kc, ociremote.WithRemoteOptions(
 					remote.WithContext(ctx),
 					remote.WithAuthFromKeychain(kc),
 				))
@@ -221,7 +225,7 @@ func (v *Validator) validatePodSpec(ctx context.Context, namespace, kind, apiVer
 					return
 				}
 
-				containerErrors := v.validateContainerImage(ctx, c.Image, namespace, field, i, kind, apiVersion, labels, ociremote.WithRemoteOptions(
+				containerErrors := v.validateContainerImage(ctx, c.Image, namespace, field, i, kind, apiVersion, labels, kc, ociremote.WithRemoteOptions(
 					remote.WithContext(ctx),
 					remote.WithAuthFromKeychain(kc),
 				))
@@ -255,7 +259,7 @@ func (v *Validator) validatePodSpec(ctx context.Context, namespace, kind, apiVer
 // or error based on the NoMatchPolicy.
 func setNoMatchingPoliciesError(ctx context.Context, image, field string, index int) *apis.FieldError {
 	// Check what the configuration is and act accordingly.
-	pcConfig := policycontrollerconfig.FromContext(ctx)
+	pcConfig := policycontrollerconfig.FromContextOrDefaults(ctx)
 
 	noMatchingPolicyError := apis.ErrGeneric("no matching policies", "image").ViaFieldIndex(field, index)
 	noMatchingPolicyError.Details = image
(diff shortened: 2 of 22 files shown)


pkg/webhook/validator_result.go
```diff
@@ -33,6 +37,10 @@ type PolicyResult struct {
 	// AuthorityMatches will have an entry for each successful Authority check
 	// on it. Key in the map is the Attestation.Name
 	AuthorityMatches map[string]AuthorityMatch `json:"authorityMatches,omitempty"`
+
+	// Config contains the Config for each of the architectures where key
+	// to the map is the architecture.
```
mattmoor, line 42: Is it the architecture or the serialized platform/architecture string?

e.g. could this be `linux/arm/v6`?

pkg/webhook/validator.go
```diff
@@ -1012,3 +1039,90 @@ func refOrFieldError(image, field string, index int) (name.Reference, *apis.Fiel
...
+		wg := new(sync.WaitGroup)
+
+		results := make(chan configFileResult, len(im.Manifests))
+		for _, manifest := range im.Manifests {
+			manifest := manifest
+			wg.Add(1)
+			go func() {
+				defer wg.Done()
+
+				newRefString := fmt.Sprintf("%s/%s@%s", ref.Context().RegistryStr(), ref.Context().RepositoryStr(), manifest.Digest)
```
mattmoor, line 1078: I think `Context()` has a `Digest(string)` method

pkg/apis/policy/v1alpha1/clusterimagepolicy_types.go
```diff
@@ -218,6 +219,13 @@ type Policy struct {
 	// ConfigMapRef defines the reference to a configMap with the policy definition.
 	// +optional
 	ConfigMapRef *ConfigMapReference `json:"configMapRef,omitempty"`
+	// FetchConfigFile controls whether ConfigFile will be fetched and made
+	// available for CIP level policy evaluation. Note that this only gets
+	// evaluated (and hence fetched) iff at least one authority matches.
+	// The ConfigFile will then be available in this format:
+	// https://github.com/opencontainers/image-spec/blob/main/config.md
+	// +optional
```
mattmoor, line 227: Should we validate that folks only set `FetchConfigFile` in the outermost policy?

pkg/apis/policy/v1alpha1/clusterimagepolicy_types.go
```diff
@@ -218,6 +219,13 @@ type Policy struct {
...
+	FetchConfigFile bool `json:"fetchConfigFile,omitempty"`
```
mattmoor, line 228: use `*bool` to distinguish between unspecified and `false`?

pkg/webhook/validator.go
mattmoor: What I mean is that `ref.Context()` is a `name.Repository` which has this method: https://github.com/google/go-containerregistry/blob/bdc946f4d482443524054e63fc56c11811858cae/pkg/name/repository.go#L114

So I think this could be: `ref.Context().Digest(manifest.Digest)`

pkg/webhook/validator.go
```diff
@@ -1012,3 +1038,104 @@ func refOrFieldError(image, field string, index int) (name.Reference, *apis.Fiel
...
+// Some examples are:
+// linux/arm64
+// linux/arm/v7
+// linux/arm/v6
+func normalizeArchitecture(cf *v1.ConfigFile) string {
+	if cf.Variant != "" {
+		return fmt.Sprintf("%s/%s/%s", cf.OS, cf.Architecture, cf.Variant)
+	}
+	return fmt.Sprintf("%s/%s", cf.OS, cf.Architecture)
+}
```
mattmoor, line 1141: Consider constructing a `v1.Platform` and using: https://github.com/google/go-containerregistry/blob/bdc946f4d482443524054e63fc56c11811858cae/pkg/v1/platform.go#L33

I think you get these for free from the multi-arch variants: https://github.com/google/go-containerregistry/blob/bdc946f4d482443524054e63fc56c11811858cae/pkg/v1/manifest.go#L49

mattmoor (dismissed):

JAORMX (commented):
why not default `fetchConfigFile` to `true`? when wouldn't one want that validation? looks quite useful to me!

pkg/apis/policy/v1alpha1/clusterimagepolicy_types.go
```diff
@@ -218,6 +219,13 @@ type Policy struct {
 	// ConfigMapRef defines the reference to a configMap with the policy definition.
 	// +optional
 	ConfigMapRef *ConfigMapReference `json:"configMapRef,omitempty"`
+	// FetchConfigFile controls whether ConfigFile will be fetched and made
+	// available for CIP level policy evaluation. Note that this only gets
+	// evaluated (and hence fetched) iff at least one authority matches.
```
hectorj2f, line 224: typo: iff

test/e2e_test_cluster_image_policy_with_config_file.sh
```diff
@@ -0,0 +1,83 @@
...
+
+echo '::group:: Remove failingClusterImagePolicy and create one that passes'
+kubectl delete -f ./test/testdata/policy-controller/e2e/cip-config-file-policy-fails.yaml
+kubectl apply -f ./test/testdata/policy-controller/e2e/cip-config-file-policy.yaml
+# allow things to propagate
+sleep 5
+echo '::endgroup::'
+
+echo '::group:: test job success'
+# We signed this above, this should work
```
hectorj2f, line 70: This comments seems incorrect.

hectorj2f: We didn't sign any image in this script

pkg/apis/policy/v1alpha1/clusterimagepolicy_types.go
```diff
@@ -218,6 +219,13 @@ type Policy struct {
...
+	// The ConfigFile will then be available in this format:
+	// https://github.com/opencontainers/image-spec/blob/main/config.md
+	// +optional
+	FetchConfigFile *bool `json:"fetchConfigFile,omitempty"`
```
hectorj2f, line 228: Could you run `make docs` ? So we get the new field in the documentation.

hectorj2f (commented):
It looks good to me, just a couple of comments

pkg/webhook/validator.go
```diff
@@ -1012,3 +1038,105 @@ func refOrFieldError(image, field string, index int) (name.Reference, *apis.Fiel
...
+		image, err := descriptor.Image()
+		if err != nil {
+			return nil, []error{fmt.Errorf("getting Image for %s: %w", ref.String(), err)}
+		}
+		cf, err := image.ConfigFile()
+		if err != nil {
+			return nil, []error{fmt.Errorf("getting ConfigFile for %s: %w", ref.String(), err)}
+		}
+		return map[string]*v1.ConfigFile{normalizeArchitecture(cf): cf}, nil
+	default:
```
hectorj2f, line 1123: Q: I am wondering if we should error if there are schema 1 images `types.DockerManifestSchema1, types.DockerManifestSchema1Signed` ? Or should we just warn ?

pkg/apis/policy/v1alpha1/clusterimagepolicy_types.go
```diff
@@ -203,7 +203,8 @@ type MatchResource struct {
 	ResourceSelector *metav1.LabelSelector `json:"selector,omitempty"`
 }
 
-// Policy specifies a policy to use for Attestation validation.
+// Policy specifies a policy to use for Attestation or the CIP validation (iff
```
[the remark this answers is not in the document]
vaikas: iff is an abbreviation for `If and only if`
https://www.merriam-webster.com/dictionary/iff

I can change it though if you want just lmk :)

pkg/apis/policy/v1alpha1/clusterimagepolicy_types.go
vaikas: Snap, totes forgot about that. We should probably add a step in GitHub actions to check for this.

pkg/webhook/validator.go
vaikas: My understanding is that they've been deprecated for looong time now and support will be removed real soon now. 
This is also how cosign deals with them:
https://github.com/sigstore/cosign/blob/main/pkg/oci/mutate/map.go#L126

vaikas:
> why not default `fetchConfigFile` to `true`? when wouldn't one want that validation? looks quite useful to me!

Great question, couple of reasons:
 * It will add some latency and unless the CIP policy actually uses them, it will be unnecessary overhead
 * There are possible error cases now (since making more outbound requests, things can fail) so unless using it in CIP policy adding failure modes for no reason seemed like a bad idea :)

hectorj2f (dismissed):

hectorj2f (approved):

PR #393, opened 2022-11-21
fix: v1beta1 version converter that ignored the field spec.policy

hectorj2f:

#### Summary

While debugging a cue policy validation, I found that the spec.policy field was not been converted, and so no validation was done based on the cue data.

#### Release Note

Fix: Policy converter was ignore spec.policy fields when converting v1beta1 policies.


vaikas (approved):
Thanks! Please add a release note, and clarify (since there's various conversions) that (unless I'm wrong) this affects v1beta1 versions only.

PR #398, opened 2022-11-28
feat: configurable ClusterImagePolicy resync period

### LINKED ISSUE #48: cosigned: add a configurable re-sync period for the image policy controller
**Description**

The reconciler needs to periodically re-sync to fetch any update on the key values especially for the keys served by KMS systems.

The value of this configurable setting could be passed as part of a new configMap with all the configuration for the image policy webhook. We'll need to set a default value.

DennyHoang:
#### Summary
* Adds policy resync period
* Closes #48 

#### Release Note
* Added --policy-resync-period arg policy-webhook for policy resync (default: 10h)

#### Documentation
https://github.com/sigstore/docs/pull/40

### DIFF
--- a/cmd/policy_webhook/main.go
+++ b/cmd/policy_webhook/main.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"flag"
 	"log"
+	"time"
 
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"knative.dev/pkg/configmap"
@@ -64,6 +65,10 @@ var (
 	//    ./config/501-policy-webhook-configurations.yaml
 	//    https://github.com/sigstore/helm-charts/blob/main/charts/policy-controller/templates/policy-webhook/policy_webhook_configurations.yaml
 	validatingWebhookName = flag.String("validating-webhook-name", "validating.clusterimagepolicy.sigstore.dev", "The name of the validating webhook configuration as well as the webhook name that is automatically configured, if exists, with different rules and client settings setting how the admission requests to be dispatched to policy-webhook.")
+
+	// policyResyncPeriod holds the interval which ClusterImagePolicies will resync
+	// This is essential for triggering a reconcile update for potentially stale KMS authorities.
+	policyResyncPeriod = flag.String("policy-resync-period", "10h", "The resync period for ClusterImagePolicies. The default is 10h.")
 )
 
 var types = map[schema.GroupVersionKind]resourcesemantics.GenericCRD{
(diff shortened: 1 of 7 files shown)


pkg/config/store.go
```diff
@@ -56,6 +58,8 @@ type PolicyControllerConfig struct {
 	NoMatchPolicy string `json:"no-match-policy"`
 	// FailOnEmptyAuthorities configures the validating webhook to allow creating CIP without a list authorities
 	FailOnEmptyAuthorities bool `json:"fail-on-empty-authorities"`
+	// PolicyResyncPeriod configures the interval for triggering ClusterImagePolicy reconcile
```
vaikas, line 61: Because this will require a controller restart and adding it here is not what this is for (as tried to indicate in the comment), I don't think it should be added here. I think it will be confusing for users to distinguish which fields take effect after a restart and which "immediately".

pkg/reconciler/clusterimagepolicy/controller.go
```diff
@@ -80,18 +83,44 @@ func NewController(
...
 		Handler: controller.HandleAll(grCb),
-	})
+	}, resyncPeriod(ctx))
+
 	return impl
 }
+
+// Obtain and return the policy-resync-period from config-policy-controller config map
+// Return default resync period when errors occur
+func resyncPeriod(ctx context.Context) time.Duration {
```
vaikas, line 106: I think it would be better to not pull this from the configmap (see earlier comment), but maybe make it a flag or env variable so it's clear that it will require a restart. You could even parse / stuff the parsed value into context and pull it from there. Other nice thing (I think) is that if you specify an invalid value, you can error out without resorting to logs.

pkg/config/store.go
DennyHoang: I was also concerned about this as well. Sure thing, I'll follow similarly to how we have "tuf-root" and "tuf-mirror" args and create another one for resync period.

cmd/policy_webhook/main.go
```diff
@@ -64,6 +65,10 @@ var (
 	//    ./config/501-policy-webhook-configurations.yaml
 	//    https://github.com/sigstore/helm-charts/blob/main/charts/policy-controller/templates/policy-webhook/policy_webhook_configurations.yaml
 	validatingWebhookName = flag.String("validating-webhook-name", "validating.clusterimagepolicy.sigstore.dev", "The name of the validating webhook configuration as well as the webhook name that is automatically configured, if exists, with different rules and client settings setting how the admission requests to be dispatched to policy-webhook.")
+
+	// policyResyncPeriod holds the interval which ClusterImagePolicies will resync
+	// This is essential for triggering a reconcile update for potentially stale KMS authorities.
+	policyResyncPeriod = flag.String("policy-resync-period", "10h", "The resync period for ClusterImagePolicies. The default is 10h.")
```
hectorj2f, line 71: Suggestion: I'll add a field to the configMap to set this value and change it without having to restart the server. We have added several new configuration settings to that config map.

vaikas: See the earlier discussion where putting it there and making it change without having to restart the server is fairly complex, unless I'm unaware of something easier to actually do this.

hectorj2f: That could be done by making some changes to knative Impl controller so the resync period could be refreshed whenever the configmap changes.
Perhaps it is too much travel for the benefit of adding that support.

hectorj2f (approved):

PR #406, opened 2022-12-03
Attach highest level resource spec to PolicyResult if so desired.

vaikas:

#### Summary
Second part of #385 , add includeSpec bool to Policy which will then attach the Spec of the resource to PolicyResult, which
then makes it available for CIP level policy evaluation.

#### Release Note
* Add includeSpec bool to spec.Policy which will then make resource Spec available for CIP level policy evaluation.


cpanato (approved):

hectorj2f (commented):
lgtm, thanks @vaikas 🚀

PR #407, opened 2022-12-04
Add includeObjectMetadata for including objectMeta in CIP policy eval.

vaikas:

#### Summary
Continuation of #406 in a way, add includeObjectMeta into spec.policy which when set to true will include the resources ObjectMeta (things like labels/annotations) in the PolicyResult that then in turn are available in evaluating with CIP.Policy iff at least one authority matches.

#### Release Note
 * Continuation of #406 in a way, add includeObjectMeta into spec.policy which when set to true will include the resources ObjectMeta (things like labels/annotations) in the PolicyResult that then in turn are available in evaluating with CIP.Policy iff at least one authority matches.

### DIFF
--- a/config/300-clusterimagepolicy.yaml
+++ b/config/300-clusterimagepolicy.yaml
@@ -82,6 +82,9 @@ spec:
                                 fetchConfigFile:
                                   description: 'FetchConfigFile controls whether ConfigFile will be fetched and made available for CIP level policy evaluation. Note that this only gets evaluated (and hence fetched) iff at least one authority matches. The ConfigFile will then be available in this format: https://github.com/opencontainers/image-spec/blob/main/config.md'
                                   type: boolean
+                                includeObjectMeta:
+                                  description: IncludeObjectMeta controls whether the ObjectMeta will be included and made available for CIP level policy evalutation. Note that this only gets evaluated iff at least one authority matches.
+                                  type: boolean
                                 includeSpec:
                                   description: IncludeSpec controls whether resource `Spec` will be included and made available for CIP level policy evaluation. Note that this only gets evaluated iff at least one authority matches. Also note that because Spec may be of a different shape depending on the resource being evaluatied (see MatchResource for filtering) you might want to configure these to match the policy file to ensure the shape of the Spec is what you expect when evaling the policy.
                                   type: boolean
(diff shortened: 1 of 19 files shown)


test/testdata/policy-controller/e2e/cip-include-objectmeta-fails.yaml
```diff
@@ -0,0 +1,29 @@
...
+  images:
+  - glob: "ghcr.io/sigstore/timestamp-server**"
+  authorities:
+  - static:
+      action: pass
+  policy:
+    includeObjectMeta: true
+    type: "cue"
+    data: |
+      objectMeta: "labels": "foo": "non-bar"
```
hectorj2f, line 29: I didn't know you can match label this way via cue. 👍🏻

pkg/webhook/validator.go
```diff
@@ -131,6 +154,7 @@ func (v *Validator) ValidatePodSpecable(ctx context.Context, wp *duckv1.WithPod)
 	// Attach the spec for down the line to be attached if it's required by
```
hectorj2f, line 154: You can ignore it, but you could include the object meta to this comment.

config/300-clusterimagepolicy.yaml
```diff
@@ -482,6 +494,12 @@ spec:
                     fetchConfigFile:
                       description: 'FetchConfigFile controls whether ConfigFile will be fetched and made available for CIP level policy evaluation. Note that this only gets evaluated (and hence fetched) iff at least one authority matches. The ConfigFile will then be available in this format: https://github.com/opencontainers/image-spec/blob/main/config.md'
                       type: boolean
+                    includeObjectMeta:
```
hectorj2f, line 497: I start to wonder if it would be better to have a single parent field and under it the objectMeta and Spec.

hectorj2f (dismissed):

config/300-clusterimagepolicy.yaml
vaikas: Yeah, hm. I don't really care one way or another :) Just one more level of indentation :) I don't see any other fields really coming here, we should pretty much have everything now, so guess if we change it to have include way you'd need to include 3 lines (include + children) to get all, where as now only 2.

pkg/webhook/validator.go
vaikas: Yup, done!

hectorj2f (approved):

PR #420, opened 2022-12-07
Cleanup: Rename `objectMeta` to `metadata` to align with K8s shape.

mattmoor:
:broom: This renamed `objectMeta:` to `metadata:` in `PolicyResult` so that the shape of what folks write for validation aligns with the shape of Kubernetes resources.

/kind cleanup

#### Release Note

🚨  THIS IS A BREAKING CHANGE 🚨, but talking to @vaikas we wanted to rip off the band-aid now before the field has meaningful adoption and cut a 0.5.1.

With this, folks can now write policies with the shape:

```yaml
metadata:
  annotations:
    foo: "bar"
spec:
  containers:
  - name: "baz"
```

### DIFF
--- a/pkg/webhook/validator.go
+++ b/pkg/webhook/validator.go
@@ -151,7 +151,7 @@ func (v *Validator) ValidatePodSpecable(ctx context.Context, wp *duckv1.WithPod)
 		return nil
 	}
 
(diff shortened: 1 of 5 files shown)


mattmoor:
cc @vaikas @hectorj2f

vaikas (approved):

PR #421, opened 2022-12-07
Add includeTypeMeta that includes TypeMeta (just like includeObjectMeta)

vaikas:

#### Summary
Add includeMetaData (just like includeObjectMeta) that includes TypeMeta (kind, apiVersion) in the PolicyResult
Also fix hack/update-codegen.sh to produce updates to both v1alpha1 and v1beta1. So, while the diffs there look large, it's because it was so out of date.

#### Release Note
 * Add includeMetaData (just like includeObjectMeta) that includes TypeMeta (kind, apiVersion) in the PolicyResult

### DIFF
--- a/pkg/webhook/validator_result.go
+++ b/pkg/webhook/validator_result.go
@@ -63,6 +63,12 @@ type PolicyResult struct {
 	// This field is only available for evaluation if
 	// CIP.Spec.Policy.IncludeObjectMeta is set to true.
 	ObjectMeta interface{} `json:"metadata,omitempty"`
+
+	// TypeMeta contains the TypeMeta for the resource that was evaluated.
+	//
+	// This field is only available for evaluation if
+	// CIP.Spec.Policy.IncludeTypeMeta is set to true.
+	TypeMeta interface{} `json:"typemeta,omitempty"`
 }
 
 // AuthorityMatch returns either Signatures (if there are no Attestations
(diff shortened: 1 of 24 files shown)


mattmoor (dismissed):

pkg/webhook/validator_result.go
```diff
@@ -63,6 +63,12 @@ type PolicyResult struct {
 	// This field is only available for evaluation if
 	// CIP.Spec.Policy.IncludeObjectMeta is set to true.
 	ObjectMeta interface{} `json:"metadata,omitempty"`
+
+	// TypeMeta contains the TypeMeta for the resource that was evaluated.
+	//
+	// This field is only available for evaluation if
+	// CIP.Spec.Policy.IncludeTypeMeta is set to true.
+	TypeMeta interface{} `json:"typemeta,omitempty"`
```
hectorj2f, line 71: is `typemeta` intentional or you want `typeMeta` similar to `authorityMatches` ?

hectorj2f (approved):

pkg/webhook/validator_result.go
vaikas: It was intentional to match metadata. Since TypeMeta is not serialized, well, it's inlined in k8s resources, figured this was ok.

mattmoor (dismissed):

PR #436, opened 2022-12-10
Plumb TrustRoot CRD through to CIP CRDs. Make TrustRoot available to webhook, clean up and refactor checkOpts logic.

### LINKED ISSUE #133: Add ctlog section to keyless examples
See https://github.com/sigstore/policy-controller/pull/118#discussion_r936025319

vaikas:
#### Summary
Wire the configmap through to the webhook so that we can start pulling the keys from it for validation. Add new fields to CIP to refer to TrustRoot so that you can specify which keys/certs should be used for validation.
Pretty big cleanup / refactoring on the validation logic to clean it up and not pass so much stuff through all over the place, unify by using cosign.CheckOpts for passing arguments, and construct them as appropriate.

Fixes #133 

#### Release Note
* add TrustRootRef to Keyless and CTLog specs so that you can specify which TrustRoot should be used for validation.


hectorj2f (commented):
Generally it looks good to me. Is it ready for a final review ?

vaikas:
I'm just adding some more tests :)

config/300-clusterimagepolicy.yaml
```diff
@@ -104,6 +104,9 @@ spec:
                         description: CTLog sets the configuration to verify the authority against a Rekor instance.
                         type: object
                         properties:
+                          trustRootRef:
+                            description: Use the Public Key from the referred TrustRoot.TLog
```
hectorj2f, line 108: I realized we are using a confusing mix of terms between CTLog and TLog imho. CTLog sets the configuration to verify the authority against a Rekor instance.For the SigstoreKey, we refer to TLog as that rekor instance. Aren't we mixing these two terms ? Or it is just me :).

docs/api-types/index.md
```diff
@@ -141,6 +141,7 @@ KeylessRef contains location of the validating certificate and the identities ag
 | url | URL defines a url to the keyless instance. | apis.URL | false |
 | identities | Identities sets a list of identities. | [][Identity](#identity) | false |
 | ca-cert | CACert sets a reference to CA certificate | [KeyRef](#keyref) | false |
+| trustRootRef | Use the Certificate Chain from the referred TrustRoot.CertificateAuthorities and TrustRoot.CTLog | string | false |
```
hectorj2f, line 144: I thought you wrote `TrustRoot.TLog` in the description. You use TrustRoot.CTLog here.

hectorj2f: I see you refer to the keyless, so you use CTLog although I will assume you refer to the rekor instance instead and so use TLog.

config/300-clusterimagepolicy.yaml
vaikas: You're not the only one confused :) That was an early screw up in the namings. I reckon we should fix this in the move to say v1beta2 or whatever the next one is that allows us to break the namings. But rather than keep the bad naming going, I tried to fix it here to be consistent with what we want it to be.

docs/api-types/index.md
vaikas: Well, this one is actually the CTLog, since it's the CA for Fulcio, and the CTLog for the Trillian that Fulcio writes to.

config/300-clusterimagepolicy.yaml
hectorj2f: I feel my confusion comes from the usage of `ctlog` here https://github.com/sigstore/policy-controller/pull/436/files#diff-83af2b17cb5361a5b3357d63c88e4965bfa07cb58917fbc0b88dd071cea39ccaR113 when we refer to a TLog resource type.

vaikas: The top level (the one on L103 is the broken name, it should be TLog). The object it points to internally is Tlog, and there we use TLog.

Spec.CTLog Should really be Spec.TLog.

vaikas: I opened https://github.com/sigstore/policy-controller/issues/442
to track this for the next API rev.

pkg/apis/policy/v1alpha1/clusterimagepolicy_types.go
```diff
@@ -161,6 +161,9 @@ type TLog struct {
 	// URL sets the url to the rekor instance (by default the public rekor.sigstore.dev)
 	// +optional
 	URL *apis.URL `json:"url,omitempty"`
+	// Use the Public Key from the referred TrustRoot.TLog
+	// +optional
+	TrustRootRef string `json:"trustRootRef,omitempty"`
```
hectorj2f, line 166: I assume this value is referred to the name of the TrustRoot, don't you prefer to use an ObjectReference instead ?

vaikas: Felt like overkill here since then we'd need to plumb in the gvk stuff as well. If you feel strongly, I can change it, just felt it would convolute the API unnecessarily.

hectorj2f: Definitely it isn't a blocker. You could use a LocalObjectReference due to the cluster scope of trustroots.

hectorj2f (approved):

PR #461, opened 2022-12-20
If TLog.url is specified, use it if trustroot does not have one + test.

vaikas:

#### Summary
 * If a trustroot is specified, and has keys but does not have URL there use the one from the TLog.URL entry. Also add a test for validating that an entry has been uploaded to Rekor.


hectorj2f (commented):
We asume any tlog to be a Rekor log. Should we document it somewhere? It is clear to me, just saying.

hectorj2f (approved):

vaikas:
> We asume any tlog to be a Rekor log. Should we document it somewhere? It is clear to me, just saying.

Yeah, so for this, I think our next version of the API will clear these up I reckon with the namings. I'll document it more thoroughly in follow ups.

Issue #460, opened 2022-12-20
Bad error message format string


mattmoor:
**Description**

These lines construct an error that wraps `nil` instead of `result.errs`:
https://github.com/sigstore/policy-controller/blob/77e766f0681f44ce1e4711c4f7a98005f8995df0/pkg/webhook/validator.go#L1204-L1205

**Version**

HEAD

cc @hectorj2f @vaikas

Issue #408, opened 2022-12-05
Add SignaturePullSecrets e2e test


hectorj2f:
**Description**

Add a signature pull secrets e2e test that uses a private registry with specific credentials to verify the usage of the signaturePullSecrets in source.

mattmoor:
We should try adding a leg that spins up KinD with these options: https://github.com/chainguard-dev/actions/blob/main/setup-kind/action.yaml#L46

> The above action logic is probably not super well tested, but it was based on https://github.com/mattmoor/mink/blob/b9148a39b2d8bbc69ca9aaf5e89a7613c0b179d8/hack/setup-kind.sh#L230-L251 which worked reasonably at some point 😅 

... and ensure that we can fetch:
1. private image/index sig/att/sbom (with `imagePullSecrets` and `signaturePullSecrets`).
2. private image/index configs (re: `fetchConfigFile`)

PR #462, opened 2022-12-20
Fix: Fix private multi-arch `fetchConfigFile`

mattmoor:
:bug: This fixes two problems fetching config files for a private multi-arch image:
1. We weren't properly surfacing errors fetching the config files.
2. We weren't passing the `remote.Option` list to the recursive call (so the `remote.WithKeychain` is dropped!)

With these changes downstream, I was able to successfully verify a private multi-arch image index.

/kind bug

Fixes: https://github.com/sigstore/policy-controller/issues/460

#### Release Note

This fixes `fetchConfigFile` for private multi-arch images, and fixes the error message handling along the same path.

#### Documentation

N/A

cc @hectorj2f @vaikas

### DIFF
--- a/pkg/webhook/validator.go
+++ b/pkg/webhook/validator.go
@@ -1187,7 +1187,7 @@ func getConfigs(ctx context.Context, ref name.Reference, options ...remote.Optio
 					return
 				}
 
-				newRefConfigs, errs := getConfigs(ctx, newRef)
+				newRefConfigs, errs := getConfigs(ctx, newRef, options...)
 				results <- configFileResult{ret: newRefConfigs, errs: errs}
 			}()
 		}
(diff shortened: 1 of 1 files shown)


vaikas:
Could we put a test together for this? I think we had some private registry tests in some of our e2e tests, would be nice to have an e2e test for this to ensure there are no regressions in the future.

mattmoor:
If we have existing private image tests, I'm happy to try and cover this there.  Do you have a pointer to where we're doing that?

vaikas:
we just set up the insecure registry
https://github.com/sigstore/policy-controller/blob/main/.github/workflows/kind-e2e-cosigned.yaml#L75

But wonder how easy it would be add auth to it and be able to use it in our tests?
https://docs.docker.com/registry/deploying/#restricting-access

Do you know how easy that would be to set up? We have this too:
https://github.com/sigstore/policy-controller/issues/408

So would be nice to get them tested. Don't want to block this change asking you to add those tests, looks fine :)

vaikas:
Or, maybe this already would do that?
https://github.com/chainguard-dev/actions/blob/main/setup-kind/action.yaml#L46

So maybe just a matter of spinning it up with that action?
https://github.com/chainguard-dev/actions/blob/main/setup-kind/action.yaml#L241

mattmoor:
@vaikas yeah, I'm pretty sure that's based on https://github.com/mattmoor/mink/blob/b9148a39b2d8bbc69ca9aaf5e89a7613c0b179d8/hack/setup-kind.sh#L230-L251 where I tested registry auth support in mink, but I'm not sure anything uses those options in our actions, and historically this stuff is finicky enough that I'm reluctant to bite off adding it here (as an example, I recall the minor version of the registry image mattered because we needed `htpasswd` to configure the auth 🤦 ).

I'll add these links to the issue you linked and it feels like it'd be a `good first issue` for someone looking to get their hands dirty upstream.

vaikas (approved):

PR #480, opened 2023-01-02
Feature: Create an interface for downstream CIP integrations.

mattmoor:
:gift: This change factors a new small library `./pkg/policy` which is intended to streamline incorporating CIP validation into downstream tooling.

For a (much) more verbose explanation see [here](https://github.com/ko-build/ko/issues/356#issuecomment-1368273840), but the general idea behind this is to allow CIP's to gate consumption of images in other contexts, for example the base images in build tools such as `ko` or `kaniko`.  The idea is to enable the tool providers to bake-in default policies for default base images, and optionally expose configuration to let users write policies to authorize base images prior to consumption.

For example, I might write the following `.ko.yaml`:
```yaml
verification:
  noMatchPolicy: deny
  policies:
  - data: |
      # inline policy
  - url: https://github.com/foo/bar/blobs/main/POLICY.yaml
```

With this library, it is likely <100 LoC to add base image policy verification to `ko`, and significantly simplifies our own `policy-tester` which has spaghetti code replicating some of this functionality.

/kind feature

#### Release Note

This adds a new `./pkg/policy` that can be used to integrate policy verification into additional tooling.

#### Documentation

N/A

### DIFF
--- a/pkg/policy/policy.go
+++ b/pkg/policy/policy.go
@@ -16,6 +16,7 @@ package policy
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"net/http"
 	"os"
@@ -29,8 +30,9 @@ type Verification struct {
 	// of the listed policies.  It allows the values: allow, deny, and warn.
 	NoMatchPolicy string `yaml:"no-match-policy,omitempty"`
 
-	// Policies specifies a collection of policies to use to cover the base
-	// images used as part of evaluation.  See "policy" below for usage.
+	// Policies specifies a set of Sources for fetching policies to use to cover
+	// images used as part of evaluation.  For more information about what each
+	// Source supports, see its usage.
 	// Policies can be nil so that we can distinguish between an explicitly
 	// specified empty list and when policies is unspecified.
 	Policies *[]Source `yaml:"policies,omitempty"`
@@ -71,65 +73,76 @@ func (v *Verification) Validate(ctx context.Context) (errs *apis.FieldError) {
 	return errs
 }
 
-func (pd *Source) Validate(ctx context.Context) (errs *apis.FieldError) {
+func (pd *Source) Validate(ctx context.Context) *apis.FieldError {
 	// Check that exactly one of the fields is set.
 	set := sets.NewString()
 	if pd.Data != "" {
 		set.Insert("data")
-		_, _, err := ParseClusterImagePolicies(ctx, pd.Data)
-		if err != nil {
-			errs = errs.Also(apis.ErrInvalidValue(err.Error(), "data"))
-		}
 	}
 	if pd.Path != "" {
 		set.Insert("path")
-		raw, err := os.ReadFile(pd.Path)
-		if err != nil {
-			errs = errs.Also(&apis.FieldError{
-				Message: err.Error(),
-				Paths:   []string{"path"},
-			})
-		} else {
-			_, _, err := ParseClusterImagePolicies(ctx, string(raw))
-			if err != nil {
-				errs = errs.Also(apis.ErrInvalidValue(err.Error(), "path"))
-			}
-		}
 	}
 	if pd.URL != "" {
 		set.Insert("url")
-		resp, err := http.Get(pd.URL)
-		if err != nil {
-			errs = errs.Also(&apis.FieldError{
-				Message: err.Error(),
-				Paths:   []string{"url"},
-			})
-		} else {
-			defer resp.Body.Close()
-			raw, err := io.ReadAll(resp.Body)
-			if err != nil {
-				errs = errs.Also(&apis.FieldError{
-					Message: err.Error(),
-					Paths:   []string{"url"},
-				})
-			} else {
-				_, _, err := ParseClusterImagePolicies(ctx, string(raw))
-				if err != nil {
-					errs = errs.Also(apis.ErrInvalidValue(err.Error(), "url"))
-				}
-			}
-		}
 	}
-
+	// This returns eagerly to avoid confusing `oneof` validation with errors
+	// along multiple paths of the oneof.
 	switch set.Len() {
 	case 0:
-		errs = errs.Also(apis.ErrMissingOneOf("data", "path", "url"))
+		return apis.ErrMissingOneOf("data", "path", "url")
 	case 1:
 		// What we want.
 	default:
 		// This will be unreachable until we add more than one thing
 		// to our oneof.
-		errs = errs.Also(apis.ErrMultipleOneOf(set.List()...))
+		return apis.ErrMultipleOneOf(set.List()...)
 	}
-	return errs
+	// We know (from the switch above) there is exactly one field name.
+	field, _ := set.PopAny()
+
+	content, err := pd.fetch(ctx)
+	if err != nil {
+		return &apis.FieldError{
+			Message: err.Error(),
+			Paths:   []string{field},
+		}
+	}
+	if _, _, err := ParseClusterImagePolicies(ctx, content); err != nil {
+		return apis.ErrInvalidValue(err.Error(), field)
+	}
+	return nil
+}
+
+func (pd *Source) fetch(ctx context.Context) (string, error) {
+	switch {
+	case pd.Data != "":
+		return pd.Data, nil
+
+	case pd.Path != "":
+		raw, err := os.ReadFile(pd.Path)
+		if err != nil {
+			return "", err
+		}
+		return string(raw), nil
+
+	case pd.URL != "":
+		req, err := http.NewRequestWithContext(ctx, http.MethodGet, pd.URL, nil)
+		if err != nil {
+			return "", err
+		}
+		resp, err := http.DefaultClient.Do(req)
+		if err != nil {
+			return "", err
+		}
+		defer resp.Body.Close()
+		raw, err := io.ReadAll(resp.Body)
+		if err != nil {
+			return "", err
+		}
+		return string(raw), nil
+
+	default:
+		// This should never happen for a validated policy.
+		return "", fmt.Errorf("unsupported policy shape: %v", pd)
+	}
 }
--- a/pkg/policy/policy.go
+++ b/pkg/policy/policy.go
@@ -1,0 +1,135 @@
+// Copyright 2023 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package policy
+
+import (
+	"context"
+	"io"
+	"net/http"
+	"os"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+	"knative.dev/pkg/apis"
+)
+
+type Verification struct {
+	// NoMatchPolicy specifies the behavior when a base image doesn't match any
+	// of the listed policies.  It allows the values: allow, deny, and warn.
+	NoMatchPolicy string `yaml:"no-match-policy,omitempty"`
+
+	// Policies specifies a collection of policies to use to cover the base
+	// images used as part of evaluation.  See "policy" below for usage.
+	// Policies can be nil so that we can distinguish between an explicitly
+	// specified empty list and when policies is unspecified.
+	Policies *[]Source `yaml:"policies,omitempty"`
+}
+
+// Source contains a set of options for specifying policies.  Exactly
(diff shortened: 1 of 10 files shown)


mattmoor:
Staging this mostly for CI for now, I want to knock off the following before this lands:
- [ ] Fold the path/url handling from policy-tester into `./pkg/policy`,
- [ ] Add `./pkg/policy/README.md` documenting how folks should integrate this.

(I'll keep updating this comment as more things come to mind)

mattmoor:
For context, this is motivated to support efforts like https://github.com/ko-build/ko/pull/918  🤞

mattmoor:
cc @vaikas @hectorj2f

pkg/policy/validate.go
```diff
@@ -0,0 +1,112 @@
...
+	// specified.
+	ErrEmptyDocument = errors.New("document is required to create policy")
+
+	// ErrUnknownType is the error returned when a type contained in the policy
+	// is unrecognized.
+	ErrUnknownType = errors.New("unknown type")
+)
+
+// Validate decodes a provided YAML document containing zero or more objects
+// and performs limited validation on them.
```
vaikas, line 42: Might want to also document that this defaults the fields for the CIPs? It's not obvious since it's buried in validate function, and I was looking for it. Others using it might wonder about it too :)

pkg/policy/policy.go
```diff
@@ -0,0 +1,135 @@
...
+			} else {
+				_, _, err := ParseClusterImagePolicies(ctx, string(raw))
+				if err != nil {
+					errs = errs.Also(apis.ErrInvalidValue(err.Error(), "url"))
+				}
+			}
+		}
+	}
+
+	switch set.Len() {
```
vaikas, line 124: This is bit clunky because if somebody specifies multiple values, we'll try to fetch/parse them, and after that will then (in addition to those errors) print this. I think it would be better to first see if the 'spec' is valid and only if it's valid (no multiples, or missing) proceed to fetch/validate them.

vaikas: Just a thought, I wonder if instead of fetching/validating in each of the if clauses above, you just fetch the bytes into something like:
```
var raw string
var field string
if pd.Data != "" {
  raw = pd.Data
  field = "data"
}

if pd.Path != "" {
  if field != "" {
      // multiple paths specified, error out
  }
field = "path"
  raw, err = os.ReadFile(...
}
```

And then after the if clauses you'd just have the now duplicated method that deals with raw alone?
```
		_, _, err := ParseClusterImagePolicies(ctx, raw)
		if err != nil {
			errs = errs.Also(apis.ErrInvalidValue(err.Error(), path))
		}
```

pkg/policy/validate.go
```diff
@@ -0,0 +1,112 @@
...
+			if warns, err = validate(ctx, uo, &v1beta1.ClusterImagePolicy{}); err != nil {
+				return
+			}
+
+		case v1alpha1.SchemeGroupVersion.WithKind("ClusterImagePolicy"):
+			if warns, err = validate(ctx, uo, &v1alpha1.ClusterImagePolicy{}); err != nil {
+				return
+			}
+
+		case corev1.SchemeGroupVersion.WithKind("Secret"):
```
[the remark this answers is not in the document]
vaikas: If the idea is that you can just cat a directory containing whatever into a single yaml, I guess I could see it, but that's not case, since if there's an unknown type we err, so yeah, just curious why the secret is special here.

pkg/policy/verifier.go
```diff
@@ -0,0 +1,210 @@
...
+// the policies backing this interface.
+type Verifier interface {
+	// Verify checks that the provided reference satisfies the backing policies.
+	Verify(context.Context, name.Reference, authn.Keychain) error
+}
+
+// WarningWriter is used to surface warning messages in a manner that
+// is customizable by callers that's suitable for their execution
+// environment.
+type WarningWriter func(string, ...interface{})
```
vaikas, line 46: Maybe my it would help me below (re: comment about surfacing warnings as typed errors below) if the arguments were documented here? What's the string, and more importantly I suppose what the interfaces here can be.

pkg/policy/verifier.go
```diff
@@ -0,0 +1,210 @@
...
+				return nil, err
+			}
+			content = string(raw)
+
+		default:
+			// This should never happen for a validated policy.
+			return nil, fmt.Errorf("unsupported policy shape: %v", p)
+		}
+
+		l, warns, err := ParseClusterImagePolicies(ctx, content)
```
vaikas, line 104: Yeah, this is kind of what I was talking about in the 'Validate' function :) But I'm wondering why we parse them above and throw away the results (in the Validate function). Given the comment below on L107 about compilation, wondering if it would make sense to to rename Validate to Compile and have it return the parsed/fetched policies instead of redoing that work here again, and any changes have to be put in both places.

pkg/policy/verifier.go
```diff
@@ -0,0 +1,210 @@
...
+	ipc *config.ImagePolicyConfig
+	ww  WarningWriter
+}
+
+// Check that impl implements Verifier
+var _ Verifier = (*impl)(nil)
+
+// Verify implements Verifier
+func (i *impl) Verify(ctx context.Context, ref name.Reference, kc authn.Keychain) error {
+	tm := getTypeMeta(ctx)
```
vaikas, line 152: We should document that some things are controlled by context fields that you can set with IncludeMeta, IncludeType because they are also part of the interface though indirectly as implemented.

pkg/policy/verifier.go
vaikas: Or put another way, wonder if there's any reason to separate Validate / Compile into separate exposed functions, or just drop Validate and have Compile alone. Just wondering since Validate is public method, it's a bit confusing since Compile already calls it and does double the work right after calling Validate. The only thing it seems to do in addition is to convert to evaluatable policies. Also since Validate does not check for duplicate policies, seems better validation to just call Compile.

vaikas (commented):
Sweet!

pkg/policy/verifier.go
```diff
@@ -0,0 +1,210 @@
...
+	matches, err := i.ipc.GetMatchingPolicies(ref.Name(), tm.Kind, tm.APIVersion, om.Labels)
+	if err != nil {
+		return err
+	}
+
+	if len(matches) == 0 {
+		switch i.verification.NoMatchPolicy {
+		case "allow":
+			return nil
+		case "warn":
```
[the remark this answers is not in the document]
mattmoor: Generally because most Go code treats normal return values and errors as mutually exclusive, so carrying warnings as wrapped errors when also continuing with normal execution is sort of awkward.  Originally I just logged it, but that presents its own "fun" (I actually need to tweak some of our logging for this to be usable, since as-is some paths spew zap logs 😞 ).

With the `apis.Validatable` pattern, the use of wrapped errors was pragmatic because:
1. It only returns an error, so it doesn't have the "other result" problem,
2. It only deals with `*apis.FieldError`, so we could add methods to mix/separate warnings/errors,
3. This kept us from breaking a several years old API! 😅 

I opted for this because you can trivially accumulate warnings with a little function like:
```
func(s string, i ...interface{}) {
   warns = append(warns, fmt.Sprintf(s, i...))
}
```
... and then we don't have to worry about plumbing anything via return values that virtually every Go programmer will hold wrong. 😅   You can also log it with `log.Printf` or treat them as errors with `log.Fatalf` (I think I use `t.Errorf` in one of the tests 🤓 ).


Happy to discuss more.  Sorry for the wall of text, just trying to dump some of my thinking here as I go through the comments.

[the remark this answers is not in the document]
vaikas: Makes sense, in other places here, you have a return signature splitting warn/errs, so maybe we could change the return signature to return:
warnings, errors

But, this is not something I'm feeling super strong about, just was thinking from the perspective of the user that wants to easily distinguish if there are errors (already handled), or just warnings. Easy part was not intuitive given the signature of the warning handler function since it only took an interface :)

pkg/policy/verifier.go
mattmoor: Ah sure, yeah this is the universal `Printf` signature 😉 

Will update in the next push.

pkg/policy/verifier.go
```diff
@@ -0,0 +1,210 @@
...
+
+	ipc *config.ImagePolicyConfig
+	ww  WarningWriter
+}
+
+// Check that impl implements Verifier
+var _ Verifier = (*impl)(nil)
+
+// Verify implements Verifier
+func (i *impl) Verify(ctx context.Context, ref name.Reference, kc authn.Keychain) error {
```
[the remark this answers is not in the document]
mattmoor: We can simply add them as a variadic argument, so that folks can opt-in if they want 👍

pkg/policy/verifier.go
mattmoor: Yeah, good point.  I added some of this after documenting most of this stuff to support `policy-tester` with policies providing `match:`.  I'll add this to the interface in the next push.

pkg/policy/validate.go
mattmoor: Will add to next push.  thanks!

pkg/policy/policy.go
```diff
@@ -0,0 +1,135 @@
...
+	"k8s.io/apimachinery/pkg/util/sets"
+	"knative.dev/pkg/apis"
+)
+
+type Verification struct {
+	// NoMatchPolicy specifies the behavior when a base image doesn't match any
+	// of the listed policies.  It allows the values: allow, deny, and warn.
+	NoMatchPolicy string `yaml:"no-match-policy,omitempty"`
+
+	// Policies specifies a collection of policies to use to cover the base
```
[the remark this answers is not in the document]
mattmoor: Yeah the "base image" thing was because I documented some of this before copying it over from `ko`, this is definitely stale.  The rest is a little confusing because I renamed this field twice.  I'll update it in the next push.

pkg/policy/policy.go
mattmoor: Yeah, I was just trying to avoid duplicating the `if` cases, I can rework this in the next push.

pkg/policy/verifier.go
mattmoor: There wasn't as much duplication when I originally went with this shape because I only had `Data`! :)

Check out the way I refactored this in the next push (tl;dr private `fetch` method to share the logic to produce `(string, error)`).

I'd sort of like to keep the `Validate` function to drop our original version of the same method downstream (as I mentioned in the `Secret` comment, but as I said there, I don't care that much, just a nice to have).

mattmoor:
Failure looks like a flake, but I can't rerun it 😞

pkg/policy/validate.go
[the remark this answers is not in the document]
vaikas: I am mainly thinking how this will start to look like assuming we start adding more things: configmaps, trustroots into the mix. That's going to make it more complicated to do more meaningful validation without starting to bleed more complexity here that will need to also be then replicated into the Compile. Secret reference support is actually a great example of that. Would the Validate unmarshal the key and make sure it's of the right type to make sure it's ok? Only to just to throw it away and redo that work in the Compile. As it stands now it doesn't do much as a validator, and actually could fail if the policy-controller is installed into a different namespace.

Not trying to be difficult here, just trying to visualize and see how this will work with more complex cases and what the division of labor is between Validate and Compile, and if Validate actually provides enough value besides making sure that the YAML can be semantically turned into objects we care about, does that make sense?

I think for CIPs it makes sense to do the Validate but for others, I am not quite convinced. At the very least, I think I'd remove the hardcoded cosign-system requirement.

Could you open a tracking issue for how much validation we really want to do besides for objects that we do not "own" (like secrets, configmaps) here vs. in Compile so we can maybe discuss it there.

pkg/policy/validate_test.go
```diff
@@ -0,0 +1,203 @@
...
+			doc: `
+apiVersion: policy.sigstore.dev/v1beta1
+kind: ClusterImagePolicy
+metadata:
+  name: blah
+spec:
+  images:
+  - glob: '*'
+`,
+			wantErr: errors.New("missing field(s): spec.authorities"),
```
vaikas, line 179: I'd prefer if we could reuse the YAMLs from the existing test/testdata for these tests rather than adding these here. But we can do this in a followup if you agree. They are used for validating the admission webhook, so would be nice to keep them in sync I think.
https://github.com/sigstore/policy-controller/tree/main/test/testdata/policy-controller
warn,invalid,valid, empty-authorities

Each of the warn, error has the expected failure baked into it, like so:
https://github.com/sigstore/policy-controller/blob/main/test/testdata/policy-controller/warn/v1beta1-missing-identity.yaml#L15

https://github.com/sigstore/policy-controller/blob/main/test/testdata/policy-controller/invalid/invalid-pubkey.yaml#L15

So thinking we would get more thorough testing and as we add more testdata files, we'll get those test cases for free.

vaikas (approved):
I reran the tests and indeed it was a flake 😠 things pass now.

I'm going to merge it for now, and we can deal with the cleanups later.

PR #482, opened 2023-01-03
Fix: Use the `apiVersion` when matching resources.

mattmoor:
:bug: I noticed that the resource matching logic was keying exclusively off of `kind` to map to a GVR, and not validating the `apiVersion` which could lead to false positive matches if other types were added to the webhook's resource match.

For example, the code today would match:
```yaml
apiVersion: mattmoor.dev/v1
kind: Deployment
```

However, the new logic does not!  I also tweaked the loop a bit to try and reduce the complexity of the boolean expressions and reduce indentation for readability.

/kind bug

#### Release Note

Fixes a bug where `ClusterImagePolicy`'s `match` might apply to resources with mismatched `apiVersion` if those types are let through the webhook.

#### Documentation

N/A

### DIFF
--- a/pkg/apis/config/image_policies.go
+++ b/pkg/apis/config/image_policies.go
@@ -23,50 +23,12 @@ import (
 	"github.com/sigstore/policy-controller/pkg/apis/glob"
 	webhookcip "github.com/sigstore/policy-controller/pkg/webhook/clusterimagepolicy"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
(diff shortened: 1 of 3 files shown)


mattmoor:
@hectorj2f @vaikas this is RFAL.  I lack the super powers to rerun the two failed legs here, but both were just 429s cloning the repo since we launch e2e tests like flying monkeys in the wizard of oz! 🤣 

cc @cpanato @priyawadhwa @dlorenc  who may be able to rerun these as well 🙏

priyawadhwa (approved):

hectorj2f (approved):

PR #498, opened 2023-01-10
update test to expected condition

k4leung4:
Signed-off-by: Kenny Leung <kleung@chainguard.dev>

#### Summary

Tweak the test to be more consistent.
It is my understanding that the [`invalid/v1beta1-empty-keyref-and-keylessref.yaml`](https://github.com/sigstore/policy-controller/blob/main/test/testdata/policy-controller/invalid/v1beta1-empty-keyref-and-keylessref.yaml) test fails with

```The ClusterImagePolicy "image-policy" is invalid: spec.authorities.keyless.identities: Invalid value: "null": spec.authorities.keyless.identities in body must be of type array: "null"```

is because of the conversion that is taking place from v1beta1 to v1alpha1.

when the webhook conversion unmarshal and re-marshal it, it caused the `identity` field to be defaulted as it doesnt have `omitempty` anymore.
so rather than an empty `keyless` field, it has an empty `identity` field which causes it to fail differently than in opensource.

before conversion
```
spec:
  images:
  - glob: image*
  authorities:
  - keyless: {}
    key: {}
```
after conversion
```
spec:
  images:
  - glob: image*
  authorities:
  - keyless:
      identities:
    key: {}
```

### DIFF
--- a/test/testdata/policy-controller/invalid/v1beta1-empty-keyref-and-keylessref.yaml
+++ b/test/testdata/policy-controller/invalid/v1beta1-empty-keyref-and-keylessref.yaml
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 ---
-# ERROR:keyless.identities: Invalid value
+# ERROR:expected exactly one, got both: spec.authorities[0].key, spec.authorities[0].keyless, spec.authorities[0].rfc3161timestamp, spec.authorities[0].static
 apiVersion: policy.sigstore.dev/v1beta1
 kind: ClusterImagePolicy
 metadata:
@@ -21,5 +21,6 @@ spec:
   images:
   - glob: image*
   authorities:
-  - keyless: {}
+  - keyless:
+      identities:
     key: {}


k4leung4:
@hectorj2f @vaikas 

I'm confused about the failure here, for the policy [v1beta1-empty-keyref-and-keylessref.yaml](https://github.com/sigstore/policy-controller/blob/main/test/testdata/policy-controller/invalid/v1beta1-empty-keyref-and-keylessref.yaml), given that both keyless and key are empty and non-nil, should it fail with the error that more than one is specified outlined here, https://github.com/sigstore/policy-controller/blob/08d0616d706f5ade7c361a5f66fd61e107b3a93c/pkg/apis/policy/v1beta1/clusterimagepolicy_validation.go#L116

but as it is, it is failing with `The ClusterImagePolicy "image-policy" is invalid: spec.authorities.keyless.identities: Invalid value: "null": spec.authorities.keyless.identities in body must be of type array: "null"`

hectorj2f:
@k4leung4 This is more an API error due to `authorities.keyless` has the identities as `omitempty` https://github.com/sigstore/policy-controller/blob/08d0616d706f5ade7c361a5f66fd61e107b3a93c/pkg/apis/policy/v1beta1/clusterimagepolicy_types.go#L179. You need to set the identities list.

If you want the other error, you need to omit `keyless: {}` and `key: {}`.

hectorj2f:
@k4leung4 We could close this PR whenever https://github.com/sigstore/policy-controller/pull/509 gets merged. That should fix your problem with the error message while keeping the API definition consistent.

Issue #479, opened 2023-01-02
REGRESSION: new failure when `ctlog` is absent


mattmoor:
**Description**

The following policy works fine with the latest released version of things, but I noticed adding unit testing at HEAD that it fails(!)

```yaml
apiVersion: policy.sigstore.dev/v1beta1
kind: ClusterImagePolicy
metadata:
  name: ko-default-base-image-policy
spec:
  images:
  - glob: cgr.dev/chainguard/static*
  authorities:
  - keyless:
      url: https://fulcio.sigstore.dev
```

I see the following error:
```
signature keyless validation failed for authority authority-0 for cgr.dev/chainguard/static@sha256:39ae0654d64cb72003216f6148e581e6d7cf239ac32325867af46666e31739d2: no matching signatures:
            checking expiry on certificate with bundle: certificate expired before signatures were entered in log: 2023-01-02T00:20:01Z is before 2023-01-02T14:37:07-08:00:
```

However, I was wondering why only certain tests failed, and this was the only one without `ctlog`.  Adding that back (since this publishes to Rekor) lets things pass:
```yaml
    ctlog:
      url: https://rekor.sigstore.dev
```

**Version**

I believe that things are fine in the latest cut release, but I assume that something about the new `TrustRoot` handling at HEAD has caused the path where `ctlog` is unspecified to be broken.

🚨  **_WE SHOULD FIX THIS BEFORE THE NEXT RELEASE_** 🚨

hectorj2f:
I believe the issue comes from https://github.com/sigstore/policy-controller/blob/main/pkg/webhook/validator.go#L1292 and returns `nil` https://github.com/sigstore/policy-controller/blob/d6ef1f37c9c634fdb2693c11f8aa91c19e76e7d8/pkg/webhook/validator.go#L1434. In a first look, I have the feeling we should default to rekor.sigstore.dev similarly to how cosign does. Otherwise we should expose a way to express a skip-tlog-verification here. We also bumped deps to use `sigstore/cosign/v2` in main, that might also cause some issues.

mattmoor:
Currently omitting `ctlog` is how we skip tlog verification 😞

vaikas:
I don't think we should be defaulting to rekor.sigstore.dev.

The intended behaviour was that if there's no ctlog specified it will skip it, but I think we're taking the offline verification path  since there's no client, but there's a public key for Rekor that's baked into the tuf root.
[...]

PR #500, opened 2023-01-11
WIP: Add a new API field to skip tlog verifications for keyless

hectorj2f:
#### Summary

This PR aims to propose a solution for #479 by adding a new field that helps us handling all the different corner cases when deciding whether to skip the tlog verification or not. 

If a `spec.authority.ctlog.url` is not defined and `skipTlogVerify=false` for a keyless authority type, this code uses the default rekor url similarly to how the cosign library does it. Otherwise, if a `spec.authority.ctlog.url` is specified then it uses these rekor settings extracted from this ctlog.url instance.

There is some additional work to be done in this PR, I'm just waiting for early feedback at this point.

### DIFF
--- a/config/300-clusterimagepolicy.yaml
+++ b/config/300-clusterimagepolicy.yaml
@@ -181,6 +181,8 @@ spec:
                                 subjectRegExp:
                                   description: SubjectRegExp specifies a regular expression to match the subject for this identity.
                                   type: string
+                          skipTlogVerify:
+                            type: boolean
                           trustRootRef:
                             description: Use the Certificate Chain from the referred TrustRoot.CertificateAuthorities and TrustRoot.CTLog
                             type: string
@@ -453,6 +455,8 @@ spec:
                                 subjectRegExp:
                                   description: SubjectRegExp specifies a regular expression to match the subject for this identity.
                                   type: string
+                          skipTlogVerify:
+                            type: boolean
                           trustRootRef:
                             description: Use the Certificate Chain from the referred TrustRoot.CertificateAuthorities and TrustRoot.CTLog
                             type: string
--- a/docs/api-types/index-v1alpha1.md
+++ b/docs/api-types/index-v1alpha1.md
@@ -260,6 +260,7 @@ KeylessRef contains location of the validating certificate and the identities ag
 | identities | Identities sets a list of identities. | [][Identity](#identity) | true |
 | ca-cert | CACert sets a reference to CA certificate | [KeyRef](#keyref) | false |
 | trustRootRef | Use the Certificate Chain from the referred TrustRoot.CertificateAuthorities and TrustRoot.CTLog | string | false |
+| skipTlogVerify |  | bool | false |
 
 [Back to TOC](#table-of-contents)
 
--- a/docs/api-types/index.md
+++ b/docs/api-types/index.md
@@ -145,6 +145,7 @@ KeylessRef contains location of the validating certificate and the identities ag
 | identities | Identities sets a list of identities. | [][Identity](#identity) | true |
 | ca-cert | CACert sets a reference to CA certificate | [KeyRef](#keyref) | false |
 | trustRootRef | Use the Certificate Chain from the referred TrustRoot.CertificateAuthorities and TrustRoot.CTLog | string | false |
+| skipTlogVerify |  | bool | false |
 
 [Back to TOC](#table-of-contents)
 
--- a/pkg/apis/policy/v1alpha1/clusterimagepolicy_types.go
+++ b/pkg/apis/policy/v1alpha1/clusterimagepolicy_types.go
@@ -185,7 +185,7 @@ type KeylessRef struct {
 	// +optional
 	TrustRootRef string `json:"trustRootRef,omitempty"`
 	// +optional
-	SkipTlogVerify *bool `json:"skipTlogVerify,omitempty"`
+	IgnoreTlog *bool `json:"skipTlogVerify,omitempty"`
 }
 
 // Attestation defines the type of attestation to validate and optionally
--- a/pkg/apis/policy/v1alpha1/clusterimagepolicy_types.go
+++ b/pkg/apis/policy/v1alpha1/clusterimagepolicy_types.go
@@ -184,6 +184,8 @@ type KeylessRef struct {
 	// Use the Certificate Chain from the referred TrustRoot.CertificateAuthorities and TrustRoot.CTLog
 	// +optional
 	TrustRootRef string `json:"trustRootRef,omitempty"`
+	// +optional
+	SkipTlogVerify *bool `json:"skipTlogVerify,omitempty"`
 }
 
 // Attestation defines the type of attestation to validate and optionally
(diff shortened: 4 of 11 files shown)


mattmoor:
How is `skipTlogVerify: false` different from passing `spec.authority.ctlog.url: https://rekor.sigstore.dev`?

I'm not sure why we need a new field either, we can already distinguish between unspecified and `ctlog: {}` since the ctlog field is a pointer and simply default `ctlog: {}` to `ctlog: { url: https://rekor.sigstore.dev }`?

Also there is a `TODO(#479)` somewhere under the new package I added where we should be able to delete a `ctlog` block. 🙏

vaikas:
RE: TODO(#479) comment, do you have a better pointer or tell me how to use grep better? :)
```
vaikas@villes-mbp policy-controller % grep -nR '479' *
go.sum:1086:golang.org/x/net v0.0.0-20210428140749-89ef3d95e781/go.mod h1:OJAsFXCWl8Ukc7SiCT/9KSuxbyM7479/AVlXFRxuMCk=
go.sum:1238:golang.org/x/tools v0.0.0-20190816200558-6889da9d5479/go.mod h1:b+2E5dAYhXwXZwtnZ6UAqBI28+e2cm9otk0dWdXHAEo=
Binary file images/intro.gif matches
Binary file main matches
test/testdata/bom-go-mod.spdx:299:PackageChecksum: SHA256: 3b17e495d6eceeb7e78bedaca087bcfc81f454c9a8dbce3ccec9071a0ecea479
Binary file tester matches
third_party/VENDOR-LICENSE/github.com/letsencrypt/boulder/goodkey/good_key.go:32:	433, 439, 443, 449, 457, 461, 463, 467, 479, 487, 491,
```

vaikas:
> How is `skipTlogVerify: false` different from passing `spec.authority.ctlog.url: https://rekor.sigstore.dev`?
> 
> I'm not sure why we need a new field either, we can already distinguish between unspecified and `ctlog: {}` since the ctlog field is a pointer and simply default `ctlog: {}` to `ctlog: { url: https://rekor.sigstore.dev }`?

The problem here has a few facets to it as I understand them, so typing them here to make sure that if I don't have the deets right, we can sort the requirements out first before making changes. 

In order to do the verification, cosign has 4 flags that one can tweak to control how the verification is done:
RekorClient, RekorPubKeys (these go hand in hand)
https://github.com/sigstore/cosign/blob/main/pkg/cosign/verify.go#L83

Offline:
https://github.com/sigstore/cosign/blob/main/pkg/cosign/verify.go#L129

SkipTLogVerify:
https://github.com/sigstore/cosign/blob/main/pkg/cosign/verify.go#L140

The original ctlog URL (which really should've been TLog, but I digress) was not enough for the various use cases. What's really required in most cases is the key to Rekor, which was implicit in the baked in TUF root and can be used to verify the Bundle.
Using the URL (which really has no direct relation to the PublicKey in the TUF) to control whether the TLog verification happens or not is not quite enough to tweak the 4 options. 

 By introducing the TrustRoot, you can now specify multiple Rekor (or just your own private different one), so that at least allows us to set the correct RekorPubKeys coupled with the URL (that eventually should also come from the TUF entry, but until it's there I think we're going to need that field still here).
 
The question is what control do we allow users for the offline and skipTLogVerify from the API. Option is bake the behaviour in to our logic, or allow them to control those options. Having said that, I'm not sure just adding the skipTLogVerify is enough. I've been noodling that if there's no URL given, that sort of kind of would imply 'offline'. But I'm not sure (need to look and haven't had time) to go see what's the interaction between passing in a nil RekorClient vs. using offline and why the cosign libraries added that extra option instead of just saying if there's no client that must mean offline.

Hope that at least clarifies a little bit on why I think the current API surface is not adequate as is to cover these cases, or let me know what I'm missing :)

hectorj2f:
> How is skipTlogVerify: false different from passing spec.authority.ctlog.url: https://rekor.sigstore.dev?
> I'm not sure why we need a new field either, we can already distinguish between unspecified and ctlog: {} since the ctlog > > field is a pointer and simply default ctlog: {} to ctlog: { url: https://rekor.sigstore.dev }?

I think @vaikas has explained the main reasons behind adding this new field to handle different scenarios. I also believe we need to align the cosign library and its different settings with the available fields for defining a policy with our API.
 
Likewise I am still wondering the benefits and whether it makes sense to support the `--offline` mode in the policy-controller.

hectorj2f:
@vaikas @mattmoor I've been checking the current`--offline` verification. I feel if we want to expose that option and `SkipTLogVerify`, we might need to add specific API fields to enable that behauvior when verifying a keyless authority.

hectorj2f:
In addition to that, I believe we'll never hit the `Offline` behauvior with the current code in main. RekorClient and RekorPubKeys will never be both `nil`, unless I am missing something 👓 .

vaikas:
FYI, there's more discussion here for various behaviours and some proposals. Mainly raising this for more context.
https://github.com/sigstore/cosign/issues/2466

hectorj2f:
@vaikas Thanks, I totally missed that discussion. I like the idea of been explicit and expecting users to define their verification mode (`--rekor-verification={offline,fallback,online,insecure}` `--tsa-verification={offline,fallback,online}`). Do we want to add support for that now ? I agree it is much better than adding multiple flags as we have now in cosign cli too.

Would we want to define this verification mode at authority level considering we have settings for tsa, rekor, fulcio, ... ?

vaikas:
I wonder if it would make sense to create an issue for this? Might be
easier to discover the decisions in the future rather than in the PR, wdyt?

Also, I think if we can balance the need for the future behaviour vs.
balancing keeping the previous release behaviour as to not break anything
without adding all the new features, that might be also an acceptable
solution. But not sure that's doable or not given the refactorings upstream?

hectorj2f:
I definitely agree this conversation needs to be moved to an issue. However our comments open the scope of this initial PR to tackle other corner cases.

vaikas:
Yes, exactly  (on expanding cope of it) :) 

That's why I'm wondering if we may be able to at least think about seeing if there's a way to provide the previous behaviour with the existing API or not. If there is, then I do wonder if we could fix it in case the API discussion takes a bit longer, and might be a breaking change, so we may want to consider bumping the API (as well as fix the other warts (like ctlog name)).

mattmoor:
@vaikas you can't find it because you removed it [here](https://github.com/sigstore/policy-controller/commit/befca08ae359832102f78d4968beb064fe84ec1e#diff-b1b1d0d360676f39ad9629e53e7227915be6509110b1e38f3e8711cd69ca0686L57) :)

vaikas:
@mattmoor I thought that was only used for validating the parsing and had no relevance for functionality being discussed here. At least it was only used to test the policy compile. So, I guess there's nothing to do for it anymore, or I missed something?

vaikas:
@Matthew Moore ***@***.***> perfect, thanks, just wanted to
make sure I wasn't missing anything :)

hectorj2f:
I am closing this PR in favor of https://github.com/sigstore/policy-controller/pull/508. We'll add new API fields in the future to handle new behaviours and future changes to the cosign library such as #505.

PR #508, opened 2023-01-16
Fix keyless behauvior when ctlog is absent

hectorj2f:

#### Release Note

Whenever you define CIP a keyless authority without a ctlog or a trustRoot, the controller will assume you are verifying your images using the sigstore ctlog `rekor.sigstore.dev`. This PR supports the old legacy behavior of keyless verification while supporting TrustRoot references. 

```
kind: ClusterImagePolicy
metadata:
  name: policy
spec:
  images:
    - glob: "**" # Your image here
  authorities:
    - name: keyatt
      keyless:
        url: "https://fulcio.sigstore.dev"
        identities:
        # Check the identity you expect to be signing SBOMs here!
        - issuerRegExp: ".*"
          subjectRegExp: ".*"
 ```
 
The controller fetches the RekorPublic keys while keeping the RekorClient nil which defaults to an offline behavior.

closes https://github.com/sigstore/policy-controller/issues/479

### DIFF
--- a/pkg/webhook/validator.go
+++ b/pkg/webhook/validator.go
@@ -38,7 +38,6 @@ import (
 	csigs "github.com/sigstore/cosign/v2/pkg/signature"
 	"github.com/sigstore/policy-controller/pkg/apis/config"
 	policyduckv1beta1 "github.com/sigstore/policy-controller/pkg/apis/duck/v1beta1"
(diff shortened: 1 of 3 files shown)


vaikas (approved):
Thanks!

Issue #458, opened 2022-12-20
Implement policy.URL


vaikas:
**Description**

I had originally lumped this into #454 but since this one is a bit different as far as how to be able to detect changes, broke into it's own issue.

We have a policy.URL field defined here:
https://github.com/sigstore/policy-controller/blob/main/pkg/apis/policy/v1alpha1/clusterimagepolicy_types.go#L224

But it's not implemented and if you try to specify it the CIP is not validated and fails.

We should implement it.
Overall should be fairly straightforward, but the one thing that's worthy of a discussion is how to handle the fact that the contents of the URL can change. Because the CIP is reconciled when there are changes to it, if the contents of the URL change, we will not be privy to it. We have a similar issue with KMS keys, but they tend to be more slow changing than say a URL (esp. when doing dev work), so using the CRD resync policy might not be the best way to handle these. We could also fetch the URL during the validation, which then introduces potentially more latency, but also failure paths. So we should think about the best way to implement this.

hectorj2f:
@vaikas We should document that any change to the kms key or url content won't be detected. We might want to find a common solution for both approaches. I feel we should add it as part of the logic added in the resync period.

vaikas:
I guess the one thing is to think about how frequently things change indeed. What I was originally worried about was the dev experience. I think one way to address the 'dev' case is that you can actually change the CIP (add a placeholder label) and force a reconcile, but it's a bit hacky. With KMS I reckon the keys are changing much less frequently so doing dev seems fine and the bigger issue is key rotation, which I think the reconcile resync handles just fine.

One thought is that using cache control headers would be probably the most natural way, but we'd still have to figure out how to honor them to get a reconcile to happen after the cache expires.

hectorj2f:
Honestly, I feel we shouldn't overcomplicate it. I see two options more k8s native in my opinion:

* Add annotations or labels to re-sync these items more often than others.
* For the url use case, I'd propose adding a `checksum` field so any change to the url needs an update of that checksum. It isn't the best experience but it is also secure.

wdyt ?

vaikas:
I agree about not overcomplicating but also want to think about the user experience is something that makes sense during operation and accounting for reliable updates when things change.
As far as k8s native, I just think making sure we keep in mind the way more well established HTTP semantics should be kept in mind so there are no surprises :) 
[...]

PR #518, opened 2023-01-20
Add support for Policy URLs

hectorj2f:
closes https://github.com/sigstore/policy-controller/issues/458

#### Summary

I am proposing a solution for setting cue-rego policies via a URL. This PR parses the url during creation time, and fetches the content of the URL to policy.Data during reconciliation.


test/testdata/policy-controller/e2e/cip-policy-from-url.yaml
```diff
@@ -0,0 +1,28 @@
...
+spec:
+  images:
+  - glob: "ghcr.io/sigstore/timestamp-server**"
+  authorities:
+  - static:
+      action: pass
+  policy:
+    fetchConfigFile: true
+    type: "cue"
+    url: "https://gist.githubusercontent.com/hectorj2f/af0d32d4be4bf2710cff76c397a14751/raw/d4dd87fffdf9624a21e62b8719e3ce8d61334ab9/policy-controller-test-fail-cue"
```
hectorj2f, line 28: TODO: change this location to a sigstore policy-controller raw file.

pkg/reconciler/clusterimagepolicy/clusterimagepolicy.go
```diff
@@ -260,6 +269,32 @@ func (r *Reconciler) inlinePolicies(ctx context.Context, cip *v1alpha1.ClusterIm
...
+			logging.FromContext(ctx).Errorf("Failed to read policy url %s: %v", cip.Spec.Policy.URL.String(), err)
+			return err
+		}
+	}
+	return nil
+}
+
+func (r *Reconciler) inlinePolicyURL(ctx context.Context, policyRef *v1alpha1.Policy) error {
+	logging.FromContext(ctx).Infof("inlining policy url %q", policyRef.URL.String())
+	resp, err := http.Get(policyRef.URL.String())
```
hectorj2f, line 284: Perhaps we can set a pre-defined timeout to avoid waiting long time for a response.

```
var client = &http.Client{
  Timeout: time.Second * 10,
}
response, _ := client.Get(policyRef.URL.String())
```

pkg/reconciler/clusterimagepolicy/clusterimagepolicy.go
```diff
@@ -260,6 +269,32 @@ func (r *Reconciler) inlinePolicies(ctx context.Context, cip *v1alpha1.ClusterIm
...
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode < 200 || resp.StatusCode > 299 {
+		return fmt.Errorf("failed to fetch content from policy url with code %q", resp.StatusCode)
+	}
+	data, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return fmt.Errorf("failed to read policy url response: %w", err)
+	}
+	policyRef.Data = string(data)
```
vaikas, line 296: I think we really should add the status field into the CIPs sooner rather than later, this one (not picking on it just reminds me of it again), that if the policy is invalid it will fail at run time. I understand that even when creating inlined CIPs and configmap refs, we need to do a better job validating the policy.

pkg/apis/policy/v1alpha1/clusterimagepolicy_validation.go
```diff
@@ -278,11 +278,23 @@ func (p *Policy) Validate(ctx context.Context) *apis.FieldError {
...
+		errs = errs.Also(apis.ErrMultipleOneOf("data", "configMapRef", "url"))
+	}
+	if (p.Data != "" && p.ConfigMapRef != nil) ||
+		(p.Data != "" && p.URL != nil) ||
+		(p.ConfigMapRef != nil && p.URL != nil) {
+		errs = errs.Also(apis.ErrMultipleOneOf("data", "configMapRef", "url"))
+	}
+	if p.URL != nil {
+		url := *p.URL
+		_, err := apis.ParseURL(url.String())
```
vaikas, line 294: I can't recall if this allows for things like relative URLs? Also, wonder if this should be required to be https?

vaikas (commented):
Took a quick pass, couple of thoughts! Thanks for working on this!

pkg/apis/policy/v1alpha1/clusterimagepolicy_validation.go
hectorj2f: I can easily improve the validation here. Yes, i like the https requirement here

pkg/reconciler/clusterimagepolicy/clusterimagepolicy.go
hectorj2f: Yes, we need to take over the status work again. I can start writing something down for this.

pkg/apis/policy/v1alpha1/clusterimagepolicy_validation.go
hectorj2f: I added some additional validations.

hectorj2f:
@vaikas I made some changes. Please, take a look when possible.

docs/api-types/index-v1alpha1.md
```diff
@@ -282,7 +283,7 @@ Policy specifies a policy to use for Attestation or the CIP validation (iff at l
 | ----- | ----------- | ------ | -------- |
 | type | Which kind of policy this is, currently only rego or cue are supported. Furthermore, only cue is tested :) | string | true |
 | data | Data contains the policy definition. | string | false |
-| url | URL to the policy data. | apis.URL | false |
+| remote | Remote defines the url to a policy. | [URL](#url) | false |
```
vaikas, line 286: Is this right? Seems like somewhere there should be the shasum field also?

vaikas: As in, shouldn't this be a pointer to #remote?

docs/api-types/index-v1alpha1.md
```diff
@@ -332,3 +333,14 @@ TLog specifies the URL to a transparency log that holds the signature and public
 | trustRootRef | Use the Public Key from the referred TrustRoot.TLog | string | false |
 
 [Back to TOC](#table-of-contents)
+
+## URL
+
+URL defines all the propierties to fetch a remote policy
```
vaikas, line 339: nit properties.

docs/api-types/index-v1alpha1.md
```diff
@@ -332,3 +333,14 @@ TLog specifies the URL to a transparency log that holds the signature and public
...
+
+| Field | Description | Scheme | Required |
+| ----- | ----------- | ------ | -------- |
+| url | URL to the policy data. | apis.URL | false |
+| sha256sum | Sha256sum defines the shasum of the policy hosted in the url. | string | false |
```
vaikas, line 344: Wonder if we should say it's the exact computation of the 'body' of the http response?

pkg/apis/policy/v1alpha1/clusterimagepolicy_types.go
```diff
@@ -214,6 +214,14 @@ type MatchResource struct {
 	ResourceSelector *metav1.LabelSelector `json:"selector,omitempty"`
 }
 
+// URL defines all the propierties to fetch a remote policy
+type URL struct {
```
vaikas, line 218: So, I don't think we should call this type URL since there's other types that are URLs. I understand it's scoped to this package, but I'd prefer to not call it URL to avoid confusion.

docs/api-types/index-v1alpha1.md
hectorj2f: yes, i like that.

pkg/apis/policy/v1alpha1/clusterimagepolicy_types.go
hectorj2f: I initially called it Remote, then I change it to avoid conflicts with the TrustRoot.Remote. I can call it RemotePolicy :D.

docs/api-types/index-v1alpha1.md
hectorj2f: fixed

docs/api-types/index-v1alpha1.md
hectorj2f: fixed

znewman01:
Is this a good idea?

I worry about an attacker just giving you the "everything passes" policy. Should we check signatures on remote policies?

hectorj2f:
@znewman01 Yes, we should check signatures, I'd like to get the policies from an OCI registry. Perhaps we could use sget logic here.
But, wouldn't it be sha256sum avoid anyone trying to change the content of a remote url without been detected by the respective sha256sum ? Obviously we should detect the identity of the policy.

znewman01:
> @znewman01 Yes, we should check signatures, I'd like to get the policies from an OCI registry. Perhaps we could use sget logic here.
> But, wouldn't it be sha256sum avoid anyone trying to change the content of a remote url without been detected by the respective sha256sum ? Obviously we should detect the identity of the policy.

D'oh, should have actually read the PR. 🤦 

Yes, you don't need signing if you're checking the sha256. But if you know the checksum, why not just inline the policy?  For size or something?

vaikas (approved):
Thanks!!! I'll rebase after merge to add these to status stuff.

PR #524, opened 2023-01-26
Improve kms key validations and error messages for awskms

### LINKED ISSUE #523: `kms` doesn't reject ARNs without `awskms://`
**Description**

Debugging a customer issue we noticed that:
```yaml
   kms: arn:...
```

... is not rejected by the validating webhook, and that the error message with `awskms://` vs. `awskms:///` (note: triple `/`!) is VERY unclear, basically what this unit test checks around "too many colons":
https://github.com/sigstore/policy-controller/blob/e386c060086fd9d68af0965258d88b8a5b69a972/pkg/apis/policy/v1alpha1/clusterimagepolicy_validation_test.go#L1307

IMO both of these should have very clear error messages that:
1. Indicate a scheme is required, and
2. Indicate that a `/` is missing instead of complaining about colons.

**Version**

0.5.2

hectorj2f:
#### Summary

closes #523 

#### Release Note

Improve kms key validations and error messages for awskms


pkg/apis/policy/common/validation.go
```diff
@@ -40,3 +50,96 @@ func ValidateOCI(oci string) error {
...
+	// Only support key or alias ARN.
+	if arn.Resource != "key" && arn.Resource != "alias" {
+		return apis.ErrInvalidValue(kms, apis.CurrentField, fmt.Sprintf("Got ARN: %+v Resource: %s", arn, arn.Resource))
+	}
+	return nil
+}
+
+func ValidateKMS(kms string) *apis.FieldError {
+	var errs *apis.FieldError
+	supportedProviders := []string{aws.ReferenceScheme, azure.ReferenceScheme, hashivault.ReferenceScheme, gcp.ReferenceScheme}
```
mattmoor, line 130: Can we make this a public global `var`, so that folks can customize the supported set?

mattmoor (dismissed):
Does this address the confusing error when someone uses `awskms://{arn}` instead of `awskms:///{arn}`?

Otherwise this looks incredibly detailed, so thanks for the better validation!

pkg/apis/policy/common/validation.go
hectorj2f: Done

hectorj2f:
> Does this address the confusing error when someone uses awskms://{arn} instead of awskms:///{arn}?

I added a specific use case for `awskms://{arn}` which is invalid but I agree we needed a better error message for this particular scenario.

mattmoor (approved):

priyawadhwa (approved):

PR #539, opened 2023-01-29
bump timeout for goreleaser to 60 minutes.

vaikas:

#### Summary
Looks like the releaser is right at the hairy edge, so bump it.
https://github.com/sigstore/policy-controller/actions/runs/4038223564/jobs/6942272664

Looks like it's been hovering there, so just bumping to 60. Previous runs:
https://github.com/sigstore/policy-controller/actions/runs/3670391147/jobs/6204895886

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -57,7 +57,7 @@ jobs:
         uses: goreleaser/goreleaser-action@8f67e590f2d095516493f017008adc464e63adb1 # v4.1.0
         with:
           version: latest
-          args: release --rm-dist
+          args: release --rm-dist --timeout 60m
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           LDFLAGS: ${{ env.GO_FLAGS }}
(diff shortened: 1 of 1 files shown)


hectorj2f (approved):
let's see if that works

.github/workflows/release.yaml
```diff
@@ -57,7 +57,7 @@ jobs:
         uses: goreleaser/goreleaser-action@8f67e590f2d095516493f017008adc464e63adb1 # v4.1.0
         with:
           version: latest
-          args: release --rm-dist
+          args: release --rm-dist --timeout 60m
```
cpanato, line 60: i would add 120m just to be sure :)

hectorj2f: we need to add a COSIGN_YES env variable. I'll do it.

cpanato: https://github.com/sigstore/policy-controller/pull/540

hectorj2f: 👍🏻

PR #541, opened 2023-01-30
fix policy controller yaml filename and update release doc

cpanato:
#### Summary
- fix policy controller yaml filename and update release doc

the release was generating the filename as `policy-controller-dirty-tag.yaml` because the `GIT_TAG` is not set and we use `GIT_VERSION` instead


hectorj2f (approved):
lgtm

PR #557, opened 2023-02-05
Improve validation when specifying a rfc3161timestamp

### LINKED ISSUE #556: Required adding key or keyless verification whenever rfc3161timestamp verification is specified
**Description**

I found our initial rfc3161timestamp implementation is treated as a first class verification, however this verification cannot be done without the presence of a key or keyless block. Therefore, we should change the CIP validations to disallow setting CIPs with only a rfc3161timestamp block.

At the moment, if a cip is created with only a rfc3161timestamp authority the cip gets created but the latest image validation fails due to the lack of any key, keyless settings to verify against.

**Version**
v0.6.0-v0.6.2

hectorj2f:
#### Summary

Nowadays, the validation allows setting an authority with only the rfc3161timestamp reference. However, we should not verify a signed image without including the expected key or keyless block. Therefore we add the required validation here.
 
Closes #556 
#### Release Note

Improve validation when specifying an authority with a RFC3161Timestamp block to require keys.

### DIFF
--- a/pkg/apis/policy/v1alpha1/clusterimagepolicy_validation.go
+++ b/pkg/apis/policy/v1alpha1/clusterimagepolicy_validation.go
@@ -91,17 +91,16 @@ func (image *ImagePattern) Validate(ctx context.Context) *apis.FieldError {
 
 func (authority *Authority) Validate(ctx context.Context) *apis.FieldError {
 	var errs *apis.FieldError
-	if authority.Key == nil && authority.Keyless == nil && authority.RFC3161Timestamp == nil && authority.Static == nil {
-		errs = errs.Also(apis.ErrMissingOneOf("key", "keyless", "rfc3161timestamp", "static"))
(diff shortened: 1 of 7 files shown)


vaikas:
Does this have any ramifications for the TrustRoot. I recall during the review for it, it was relaxed to allow just the timestampauthorities without say certificateauthorities. Is that still the case? Just wondering how that should be changed.
https://github.com/sigstore/policy-controller/blob/main/pkg/apis/policy/v1alpha1/trustroot_validation.go#L108

hectorj2f:
For a general purpose, I feel this [validation](https://github.com/sigstore/policy-controller/blob/main/pkg/apis/policy/v1alpha1/trustroot_validation.go#L108) is fine. You could validate images with a combination of a timestamp authority, a fulcio sigstore keyless and a ctlog without setting a specific certificate authority.

vaikas (approved):

PR #582, opened 2023-02-13
add validation for secretRef.namespace value

hectorj2f:

#### Summary

Throw an error whenever the secretRef sets a namespace that isn't equal to the namespace where policy-controller was deployed.

#### Release Note

Add validation for the secretRef.namespace field.


vaikas (approved):
Thanks! Looks like there's one small whitespace issue.

vaikas (approved):

PR #596, opened 2023-02-18
Cleanup: Move where we restrict resource types.

mattmoor:
:broom: This change moves a bunch of shared constants from `v1alpha1`/`v1beta1` into `common`

:broom: This also changes the behavior of `ValidResourceNames` slightly.  Rather than having the CIP validation bake in a fixed set of resources in the library (in a private variable), the binary entrypoint defines these close to where we set up the resource map, since these must be consistent!

/kind cleanup

#### Release Note

Enable the controller entrypoint to control the supported resource set validated in the webhook.

### DIFF
--- a/cmd/policy_webhook/main.go
+++ b/cmd/policy_webhook/main.go
@@ -22,6 +22,7 @@ import (
 	"time"
 
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"knative.dev/pkg/configmap"
 	"knative.dev/pkg/controller"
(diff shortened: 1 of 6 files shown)


mattmoor:
As context, I was trying to play around with a policy on a Knative service in policy tester, and ran into this.  This feels like something the binary entrypoint should have more control over

hectorj2f (approved):

hectorj2f:
I agree, it makes sense.

mattmoor:
@hectorj2f anything else we should think about here, or is this good to go?

hectorj2f:
LGTM

PR #602, opened 2023-02-21
Bump to Cosign v2.0.0 rc.3

hectorj2f:
#### Summary

Adapt code to changes in cosign v2.0.0-rc.3 where several things changed, especially the predicate type verifications.

#### Release Note

* Use cosign v2.0.0-rc.3
    * BREAKING: Any attestation verification will now use the RFC 3986 compliant URIs (adds https://) so that the predicate types must be compliant with the in-toto specification, e.g. https://cosign.sigstore.dev/attestation/vuln/v1.
    * BREAKING: This new cosign version also removes the CosignPredicate envelope that wraps the predicates of SPDX and CycloneDX attestations. Therefore, policies must specify the explicit predicate type, e.g. https://spdx.dev/Document. In addition to that, any occurrence of predicate Data block should be now removed from any policy. As an example, a predicate such as `predicate: Data: spdxVersion: ...` should be replaced by `predicate: spdxVersion: ...`, e.g. `predicate: Data: specVersion: ...` --> `predicate: specVersion: ...`.


mattmoor:
@vaikas @hectorj2f WDYT about deprecating and starting to warn when folks use the short-name version of the predicate type?

I'd love it if we could get folks using explicit predicate types, since the short names have been so fraught.

Thanks for bumping this @hectorj2f 👍

hectorj2f:
Warning always looks like a good approach announcing for a future deprecation. Yes, it makes sense to slowly transition to  strict predicate types. I'm wondering if we could start adding a warning in cosign (too) for these short names.

vaikas:
I think this has a few breaking changes that we should make sure make it into the release notes.

vaikas:
There's also this:
https://github.com/sigstore/cosign/pull/2718

Which means that all the policies written against spdx/cyclonedx will not work as far as I can tell?

I'd go through the PRs from the last and make sure there weren't others.

hectorj2f:
@vaikas Thanks for pointing that out. In my opinion, the other PRs are not affecting our verifications if I am not mistaken.

vaikas:
@hectorj2f thanks for checking. I would word it that you have to rewrite your policy, not only that you must specify the predicatetype, but the extra 'indentation' of `Data` has been removed, so policies that used them will need to be changed:
https://github.com/sigstore/cosign/pull/2718/files#diff-9e762d0df1502ee31a1de1a3bc67be6c2707ec3e6117be3147907a60ef5fb854L263

hectorj2f:
@vaikas I feel we should highlight these changes when documenting (docs/README) the new version.

vaikas (approved):

vaikas:
@hectorj2f we can certainly create a 'changelog.MD' or something like that in addition to clear commit messages that impact users, but for folks that are looking to upgrade to a new version are accustomed to seeing the breaking changes when they look at the releases.

hectorj2f:
Thanks for the reviews!

PR #614, opened 2023-02-23
bump cosign to v2.0.0 release

k4leung4:
#### Summary

bump cosign to v2.0.0 release


hectorj2f:
We need to adapt the call to the new function calls, it might also need to adapt it to the error messages too.

vaikas (approved):
Thank you @k4leung4

PR #637, opened 2023-03-06
Use default cosign-installer version

hectorj2f:
#### Summary

Stop setting the cosign version in our gh actions, and use the default cosign version from the installer.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -36,8 +36,6 @@ jobs:
       - uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # v3.3.0
 
       - uses: sigstore/cosign-installer@c3667d99424e7e6047999fb6246c0da843953c65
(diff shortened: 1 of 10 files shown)


.github/workflows/build.yaml
```diff
@@ -36,8 +36,6 @@ jobs:
       - uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # v3.3.0
 
       - uses: sigstore/cosign-installer@c3667d99424e7e6047999fb6246c0da843953c65
```
cpanato, line 38: can you also check if this hash is from the latest release?

hectorj2f: yes it is

hectorj2f: oh, it got merged a new PR 23min ago.

cpanato: thanks

Issue #631, opened 2023-03-01
Combine the two controller pods into a single pod


hectorj2f:
**Description**

Nowadays, the policy-controller deploys two different controllers one for the cluster image policy and another for the validations. We want to run both controllers into a single pod.

PR #717, opened 2023-04-11
Combine the two controllers into a single pod

hectorj2f:
#### Summary

Fixes: https://github.com/sigstore/policy-controller/issues/631

#### Documentation

Unify our two main controllers into a single pod


hectorj2f:
Locally works but it seems there is something going on on the github actions.

anderssonw:
@hectorj2f Let me know if you need any particular help here. Would love to get this through before we do further work on our use of the controller, as this greatly reduces noices :)

hectorj2f:
Thanks @anderssonw! I'm still investigating the issue but I have an idea of what is happening.

hectorj2f:
This PR needs this bug fix to work https://github.com/knative/pkg/pull/2731.

config/501-policy-webhook-configurations.yaml
```diff
@@ -36,16 +36,10 @@ webhooks:
...
     matchPolicy: Equivalent
     name: validating.clusterimagepolicy.sigstore.dev
     sideEffects: None
 ---
-apiVersion: v1
-kind: Secret
-metadata:
-  name: policy-webhook-certs
-  namespace: cosign-system
-# The data is populated at install time.
```
cpanato, line 51: this is not needed anymore?

config/webhook.yaml
```diff
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
+---
```
cpanato, line 14: do we need this extra ticks?

cpanato (dismissed):
lgtm

just a few nit comments

config/501-policy-webhook-configurations.yaml
hectorj2f: Yes, we are getting rid and unifying all the needs to one single pod.

hectorj2f:
> but the title is in a single container instead of single pod right?

@cpanato  It is in a single pod. We had two pods before.

config/webhook.yaml
hectorj2f: We need it to construct the yaml file with all the different yamls.

vaikas (approved):

PR #732, opened 2023-04-13
Add policy-controller annotations

### LINKED ISSUE #611: Add annotations to objects admitted by policy-controller
**Description**

It would be awesome to see the results of policy-controller admissions recorded on affected objects.

There's a great example from [Tekton Chains](https://github.com/tektoncd/chains/blob/main/docs/tutorials/signed-provenance-tutorial.md#verifying-the-image-and-attestation) where they record helpful metadata:
```
kubectl get tr [TASKRUN_NAME] -o json | jq -r .metadata.annotations

{
  "chains.tekton.dev/signed": "true",
  ...
}
```

The things that seem most useful to me from the policy-controller side would be an annotation referencing the admitting or denying policy, the status, and ideally, enough details to understand the reasoning behind the policy decision.

elfotografo007:
#### Summary

Add annotations to a resource with the validation results.
Closes #611 

#### Example Output
Given the following policy, and enabling the policy-controller in the namespace `my-secure-namespace`:
```
kind: ClusterImagePolicy
metadata:
  name: image-policy
spec:
  images:
  - glob: "gcr.io/projectsigstore/cosign*"
  authorities:
  - name: official-cosign-key
    key:
      data: |
        -----BEGIN PUBLIC KEY-----
        MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEhyQCx0E9wQWSFI9ULGwy3BuRklnt
        IqozONbbdbqz11hlRJy9c7SG+hdcFl9jE9uE/dwtuwU2MqU9T/cN0YkWww==
        -----END PUBLIC KEY-----
```

Creating a pod with `kubectl run cosign --image=gcr.io/projectsigstore/cosign:v1.2.1 --dry-run=server -n my-secure-namespace -o yaml` will generate the following annotation:
```
apiVersion: v1
kind: Pod
metadata:
  annotations:
    policy.sigstore.dev/policy-controller-results: '{"containerResults":[{"index":0,"name":"cosign","image":"gcr.io/projectsigstore/cosign@sha256:68801416e6ae0a48820baa3f071146d18846d8cd26ca8ec3a1e87fca8a735498","field":"containers","result":"allow","resultMsg":"Validated 1 policies for image gcr.io/projectsigstore/cosign@sha256:68801416e6ae0a48820baa3f071146d18846d8cd26ca8ec3a1e87fca8a735498","policyResults":{"image-policy":{"authorityMatches":{"official-cosign-key":{"signatures":[{"id":"fe11e4e3fc2d0b19518341289fabcb3d36102308bf8b5d773a2eefd1251df086"},{"id":"9da9bce3022befe4f55e9dea0680a2ae48141570c98df6c541c557fa533fe2f1"}]}}}}}]}'
  creationTimestamp: "2023-04-13T16:56:25Z"
  labels:
    run: cosign
  name: cosign
  namespace: my-secure-namespace
...
```

The annotations are in JSON format.

#### Release Note

Added annotations to validated resources.
#### Documentation

[Docs PR](https://github.com/sigstore/docs/pull/127)


cmd/webhook/main.go
```diff
@@ -202,6 +202,11 @@ func NewValidatingAdmissionController(ctx context.Context, cmw configmap.Watcher
 }
 
 func NewMutatingAdmissionController(ctx context.Context, cmw configmap.Watcher) *controller.Impl {
+	store := config.NewStore(logging.FromContext(ctx).Named("config-store"))
+	store.WatchConfigs(cmw)
+	policyControllerConfigStore := policycontrollerconfig.NewStore(logging.FromContext(ctx).Named("config-policy-controller"))
```
hectorj2f, line 207: Let's add an option to the configMap `config-policy-controller` to enable this feature on demand.

elfotografo007: Sure, I'll update the PR soon

pkg/config/store.go
```diff
@@ -43,6 +43,8 @@ const (
 	NoMatchPolicyKey = "no-match-policy"
 
 	FailOnEmptyAuthorities = "fail-on-empty-authorities"
+
+	AnnotateResultsKey = "annotate-validation-results"
```
hectorj2f, line 47: nit: `AnnotateValidationResultsKey`

hectorj2f (commented):
We need to add a new github action worklflow to test this functionality on a running cluster similar to what we did with kind-cluster-image-policy-tsa.yaml or kind-cluster-image-policy-no-tuf.

pkg/webhook/validator.go
```diff
@@ -1063,6 +1087,332 @@ func (v *Validator) resolvePodSpec(ctx context.Context, ps *corev1.PodSpec, opt
...
+	}
+
+	ns := getNamespace(ctx, p.Namespace)
+	opt := k8schain.Options{
+		Namespace:          ns,
+		ServiceAccountName: p.Spec.ServiceAccountName,
+		ImagePullSecrets:   imagePullSecrets,
+	}
+
+	v.annotatePodSpec(ctx, ns, p.Kind, p.APIVersion, &p.ObjectMeta, &p.Spec, opt)
```
hectorj2f, line 1117: We could get from the context whether annotate-results- configuration is enabled or not.

hectorj2f:
@elfotografo007 Let me know if you have any questions or need help.

cmd/webhook/main.go
```diff
@@ -218,10 +223,12 @@ func NewMutatingAdmissionController(ctx context.Context, cmw configmap.Watcher)
...
 		func(ctx context.Context) context.Context {
 			ctx = context.WithValue(ctx, kubeclient.Key{}, kc)
-			ctx = policyduckv1beta1.WithPodScalableDefaulter(ctx, validator.ResolvePodScalable)
-			ctx = duckv1.WithPodDefaulter(ctx, validator.ResolvePod)
-			ctx = duckv1.WithPodSpecDefaulter(ctx, validator.ResolvePodSpecable)
-			ctx = duckv1.WithCronJobDefaulter(ctx, validator.ResolveCronJob)
+			ctx = store.ToContext(ctx)
+			ctx = policyControllerConfigStore.ToContext(ctx)
+			ctx = policyduckv1beta1.WithPodScalableDefaulter(ctx, validator.PodScalableDefaulter)
+			ctx = duckv1.WithPodDefaulter(ctx, validator.PodDefaulter)
```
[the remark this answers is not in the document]
elfotografo007: I didn't want to couple the annotations with `ResolvePod`, so we can change or get rid of the annotations easily. This also prevents modifying the `ResolvePod` tests, which may become a little unmanageable.

pkg/webhook/validator.go
elfotografo007: Sorry, I am not sure what you are referring to. I make the check from the context in one place in line 1206.

elfotografo007:
@hectorj2f Can you re-run the tests? I believe they failed due to a GH outage.

hectorj2f:
@elfotografo007 Done

elfotografo007:
Thanks, @hectorj2f Everything is passing now 😃

cmd/webhook/main.go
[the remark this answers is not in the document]
hectorj2f: I hope we don't need to get rid of the annotations already :). I think it is unnecessary to maintain two similar functions of defaulting our pod resolvers, and instead add the annotations only when needed.

pkg/webhook/validator.go
hectorj2f: Oh! This comment was related to this other one https://github.com/sigstore/policy-controller/pull/732#discussion_r1197431379. I believe we could simply get from the context whether our system had enabled the annotations on resources or not.That'd reduce the amount of code for this feature while keeping the behaviour configurable.

cmd/webhook/main.go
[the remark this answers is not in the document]
hectorj2f: We could have a call so I can help you to turn the code into a more manageable one.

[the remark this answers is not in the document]
hectorj2f: @vaikas wdyt ?

vaikas:
Just to make sure I understand, does this mean that we run through the policy evaluations twice if the annotations are enabled? Once during defaulting (and that's when the annotation gets applied), and again once all the defaultings are done and the actual admission check happens we run through the same evaluations again?
If that's the case it creates 2x the load on registries, but more importantly there is a chance that what gets written to the annotation is not what gets evaluated.
Just wanted to check my understanding.

elfotografo007:
You are totally right, everything you are mentioning is what is happening. Is there a way to evaluate and create annotations at the same time?

vaikas:
> You are totally right, everything you are mentioning is what is happening. Is there a way to evaluate and create annotations at the same time?

I don't believe so in a sense that you want to do the admission / create annotations at the same time. I thought there was a way to return more information along the admission webhook response. It really depends on what behaviour we want to provide.

One option that I was thinking was using the 'AuditAnnotations' from here:
https://pkg.go.dev/k8s.io/kubernetes/pkg/apis/admission#AdmissionResponse

evenh:
Looking forward to this feature 🤞🏻

hectorj2f:
@elfotografo007 Do you have any update about this one ? Are you still working on these changes ? :)

ElementTech:
@elfotografo007 @hectorj2f Did someone look at [this part](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#response) of the admission controller documentation? Where you could do a JSON Patch to an admitted object. I really need this feature so I tried looking in the code where a response is returned but didn't manage to understand.

Basically Base64 encoding this `[{"op": "add", "path": "/spec/replicas", "value": 3}]` for example:

```
{
  "apiVersion": "admission.k8s.io/v1",
  "kind": "AdmissionReview",
  "response": {
    "uid": "<value from request.uid>",
    "allowed": true,
    "patchType": "JSONPatch",
    "patch": "W3sib3AiOiAiYWRkIiwgInBhdGgiOiAiL3NwZWMvcmVwbGljYXMiLCAidmFsdWUiOiAzfV0="
  }
}
```

hectorj2f:
@ElementTech no, we didn't. How do you want to use it?

PR #809, opened 2023-05-30
Fix: Have the policy-tester library check policy result.

mattmoor:
:bug: The policy tester library blindly checked the errors, but with multiple authorities we might have errors, but still successfully evaluated the policy.

Here is the comment from the ~equivalent call to `ValidatePolicy(` from `validator.go` for comparison:

```go
			switch {
			// Return AuthorityMatches before errors, since even if there
			// are errors, if there are 0 or more authorities that match,
			// it will pass the Policy. Of course, a CIP level policy can
			// override this behaviour, but that has been checked above and
			// if it failed, it will nil out the policyResult.
			case result.policyResult != nil:
				policyResults[result.name] = result.policyResult
			case len(result.errors) > 0:
				ret[result.name] = append(ret[result.name], result.errors...)
			default:
				ret[result.name] = append(ret[result.name], fmt.Errorf("failed to process policy: %s", result.name))
			}

```

... with this I was able to successfully run the policy tester using a policy that had a separate authority for SLSA `v0.2` and `v1` predicate types without it failing.

/kind bug

#### Release Note

The policy tester library now properly treats authorities as an `OR` instead of erroring if any of them surface errors.

### DIFF
--- a/pkg/policy/verifier.go
+++ b/pkg/policy/verifier.go
@@ -157,18 +157,23 @@ func (i *impl) Verify(ctx context.Context, ref name.Reference, kc authn.Keychain
 	opts = append(opts, ociremote.WithRemoteOptions(remote.WithAuthFromKeychain(kc)))
 
 	for _, p := range matches {
-		_, errs := webhook.ValidatePolicy(ctx, "" /* namespace */, ref, p, kc, opts...)
-		for _, err := range errs {
-			var fe *apis.FieldError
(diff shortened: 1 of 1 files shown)


mattmoor:
cc @vaikas @hectorj2f

vaikas (approved):

hectorj2f:
Thanks @mattmoor

PR #826, opened 2023-06-07
Generate unique PolicyResult.ID for the same image.

### LINKED ISSUE #825: Signature IDs aren't unique for the same image
**Description**

`PolicyResult.ID` is intended to be unique: https://github.com/sigstore/policy-controller/blob/bb9d59ba344de291fd076259f945c452e4b5564a/pkg/webhook/validator_result.go#L94-L96

Currently, when a PolicySignature is created, we try to create a unique identifier by getting the digest of the signature - https://github.com/sigstore/policy-controller/blob/bb9d59ba344de291fd076259f945c452e4b5564a/pkg/webhook/validator.go#L645

But non-intuitively, this doesn't give you the digest of the signature itself, this gives you [the digest of the content that was signed](https://github.com/sigstore/cosign/blob/871448050b924a7946ebe47678f23aae09ef432d/pkg/oci/static/signature.go#L188-L191) (the signature / cert are included in layer annotations). This means that any signatures for the same image will have the same `PolicySignature.ID`

We should modify this to generate a digest of the signature itself.

**Version**

bb9d59ba344de291fd076259f945c452e4b5564a (HEAD)

wlynch:
#### Summary

This changes how IDs are created to use the signature digest rather than layer digest. Because of how layers are created, this was actually a digest of the signed content, which means that multiple signatures over the same content would yield the same ID.

Instead, this uses a checksum of the signature, which should be unique enough for both content + key.
Assuming PolicyResult.ID is an opaque detail for users, this shouldn't be an invasive change, since both produce sha256 checksums.

Fixes #825 

#### Release Note

Fixed issue where multiple signatures on the same content would generate the same PolicyResult IDs.


hectorj2f (approved):
lgtm

vaikas (approved):
Thank you!

wlynch:
Converting this to draft - we probably need to go a step further since the certs could also be different. 😭

hectorj2f (approved):

PR #884, opened 2023-07-06
Add tooling for a local dev environment

malancas:
#### Summary

This pull request adds a small CLI tool, currently called `localk8s`, that allows developers to easily spin up and spin down a local Kind cluster to develop the policy controller in. This tool can be expanded in another PR to support pushing images to the local Kind registry.

I have a bash script version of this work that spins up a Kind cluster but thought that having a CLI tool like this could allow developers to add support for additional configurations more easily than a bash script.

Follow up work includes:
- Adding configuration that allows users to upload images to the local Kind registry instead of an online registry


malancas:
cc @vaikas @hectorj2f

vaikas:
Yeah, we do have quite a few bash scripts that do this, if you do find this easier to work with, great! As far as the local kind registry, in the scaffolding (and few other places 🤣 ) do spin it up:
```
https://github.com/sigstore/scaffolding/blob/main/hack/setup-kind.sh#L261
```

vaikas (approved):

Makefile
```diff
@@ -93,6 +93,11 @@ policy-controller:
 policy-tester:
 	CGO_ENABLED=0 go build -trimpath -ldflags "$(LDFLAGS)" -o $@ ./cmd/tester
 
+## Build localk8s binary
+.PHONY: localk8s
```
hectorj2f, line 97: Suggestion\nit: I would name it `local-dev` as we assume policy-controller just rans on k8s.

hectorj2f (dismissed):
Lgtm, I would just omit using k8s prefixes and name it local-dev or similar.

PR #894, opened 2023-07-14
Add support for local registry use to the local-dev tooling

malancas:
#### Summary

Adds support for using the local Kind registry. This pull request also adds a new `k8s-version` flag. If supplied, this option determines which Kind image is used when creating the Kind cluster.

cc @hectorj2f @vaikas


hack/local-dev.sh
```diff
@@ -0,0 +1,130 @@
+#!/bin/bash
+
+# Copyright 2022 The Sigstore Authors
```
hectorj2f, line 3: Copyright 2023

hack/local-dev.sh
```diff
@@ -0,0 +1,130 @@
...
+# The script accepts the following optional arguments:
+# 1. cluster-name - default value is 'policy-controller-demo'
+# 2. k8s-version - default value is 'v1.24.x'
+# 3. registry-url - if no value is provided, the local Kind registry is used
+#
+
+set -o errexit
+
+CLUSTER_NAME="policy-controller-demo"
+K8S_VERSION="v1.24.x"
```
hectorj2f, line 26: Could we default to 1.26.X ?

hack/local-dev.sh
```diff
@@ -0,0 +1,130 @@
...
+fi
+
+export GIT_HASH=$(git rev-parse HEAD)
+export GIT_VERSION=$(git describe --tags --always --dirty)
+
+CONFIG_FILES=$(find config -name "*.yaml" ! -name 'kustomization.yaml' | sort)
+
+for i in ${CONFIG_FILES[@]}
+do
+  ko apply -f $i
```
hectorj2f, line 129: Why don't you run `make ko-apply` instead ?

cpanato (commented):
Personally i think we should stay using go instead of bash

hectorj2f:
I am okay with any approach as long as it is easy to mantain in a longer term.i don't have strong preferences here.

malancas:
Thanks for the feedback, I was able to implement more with the Go CLI. I'll push an update soon that includes it.

cpanato:
cool!

cmd/local-dev/setup.go
```diff
@@ -17,21 +17,46 @@ package main
...
+  [plugins."io.containerd.grpc.v1.cri".registry.mirrors."%s:%d"]
+    endpoint = ["http://%s:%d"]
+`
+
 func addSetupFlags(cmd *cobra.Command) {
 	cmd.Flags().String("cluster-name", "policy-controller-demo", "name of the dev policy controller cluster")
-	cmd.Flags().String("ko-docker-repo", "", "name of the Ko Docker repository to use")
-	cmd.MarkFlagRequired("ko-docker-repo") //nolint:errcheck
+	cmd.Flags().String("registry-url", "registry.local", "URL of the Ko Docker registry to use. If no registry is provided, the local Kind registry will be used")
+	cmd.Flags().String("k8s-version", "v1.26.x", "name of the Ko Docker repository to use")
```
hectorj2f, line 59: If we use `v1.x.x` format, we should explain the other valid values. It isn't clear to me which is a valid k8s version. Perhaps we should use regexp to valid the chosen version.

cmd/local-dev/setup.go
```diff
@@ -121,6 +171,68 @@ func setup() {
...
+	}
+
+	fmt.Println("Connecting network between kind with local registry ...")
+
+	return cli.NetworkConnect(context.Background(), "kind", localRegistryName, nil)
+}
+
+func getKindImage(k8sVersion string) string {
+	switch k8sVersion {
+	case "v1.23.x":
```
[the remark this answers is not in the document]
malancas: Sure, I'll look into handling this with regexp.

hectorj2f (approved):

malancas:
thanks for the review @hectorj2f, would you mind merging the PR? I don't have access to merge it

cpanato (approved):

PR #1158, opened 2023-12-19
refactor(refOrFieldError): Update error text

Minivolk02:
Hey!
I think by mistake no image digest is not entirely clear what is meant, so it’s better to add to the message that perhaps this image does not exist at all


cpanato:
@Minivolk02 please sign the DCO

Issue #1297, opened 2024-03-07
Convert SigstoreKeys to TrustedRoot from protobuf-specs


codysoyland:
**Description**

Sigstore's TUF layout historically has used a collection of [target files], along with [custom metadata] to represent the valid key material for Fulcio, Rekor, and the CT Log. More recently, a new specification has been accepted, the [`TrustedRoot`] data structure, which combines these keys into a single message/filetype, independent of TUF, and available as a target in the Sigstore root TUF: [trusted_root.json].

This helps with simplicity and portability, and adds explicit date ranges so that keys may be rotated. This format is in broad use across language-specific Sigstore clients, such as `sigstore-js`, [`sigstore-go`], and `sigstore-python` (et al.).

As part of an effort to modernize the Policy Controller, I propose that [`SigstoreKeys`] be replaced with the `TrustedRoot` type. This is not my idea: the comments indicate that @vaikas intended to follow up with this.

This will pave the way for integration with `sigstore-go` to verify the [Sigstore Bundle], as part of a broader initiative to use Sigstore Bundles as the common format for encapsulating attestations, signatures, and related key/transparency material, as described in [Sigstore Bundle as OCI Artifact].

[target files]: https://github.com/sigstore/root-signing/tree/main/targets
[custom metadata]: https://github.com/sigstore/root-signing/blob/main/repository/repository/targets.json
[`TrustedRoot`]: https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_trustroot.proto#L101
[trusted_root.json]: https://github.com/sigstore/root-signing/blob/main/targets/trusted_root.json
['sigstore-go]: https://github.com/sigstore/sigstore-go/
[`SigstoreKeys`]: https://github.com/sigstore/policy-controller/blob/main/pkg/apis/config/sigstore_keys.go#L84-L102
[Sigstore Bundle]: https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_bundle.proto#L111
[Sigstore Bundle as OCI Artifact]: https://github.com/sigstore/cosign/issues/3577

PR #1155, opened 2023-12-17
re-create TUF root for testing purposes

hectorj2f:
closes https://github.com/sigstore/policy-controller/issues/1154

#### Summary

We need to re-create the TUF root which expired very recently breaking our CI builds.

PR #1303, opened 2024-03-12
Convert SigstoreKeys to protobuf-specs type

codysoyland:
#### Summary

This PR converts [`SigstoreKeys`] into the [`TrustedRoot`](https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_trustroot.proto#L101) type from `protobuf-specs`.

Additionally, this updates the TUF fetching logic to use `trusted_root.json` to fetch the trusted key material if it exists, and falls back to using the custom metadata and individual target files if that file is not included in given TUF repo.

This is the first step in modifying policy-controller to support bundle verification with [`sigstore-go`] using the OCI storage scheme detailed in [Sigstore Bundle as OCI Artifact].

Fixes https://github.com/sigstore/policy-controller/issues/1297

[`SigstoreKeys`]: https://github.com/sigstore/policy-controller/blob/main/pkg/apis/config/sigstore_keys.go#L84-L102
[`sigstore-go`]: https://github.com/sigstore/sigstore-go
[Sigstore Bundle as OCI Artifact]: https://github.com/sigstore/cosign/issues/3577


hectorj2f:
@codysoyland Let me know when it is ready for a review. Good work :)!

codysoyland:
> @codysoyland Let me know when it is ready for a review. Good work :)!

Thanks @hectorj2f! A few questions if you don't mind:

- The serialized test data in the reconciler tests (e.g. `marshalledEntry` and `marshalledEntryFromMirrorFS`) need to be regenerated any time I make a change to `SigstoreKeys`. I was thinking about replacing these with JSON files using `go:embed` and writing a script (added to `/hack`) to generate them, if you'd be open to that?
- I have the same question about the `validRepository`/`rootJSON` values in the tests (serialized TUF repos). You don't currently have a script to do this, do you? I see you [updated them somewhat recently](https://github.com/sigstore/policy-controller/pull/1155). If you have any notes, that would be helpful -- I'm assuming you used the go-tuf CLI?
- I don't want this patch to break anyone's existing K8s setup, but it does change the structure of the cached keys in the ConfigMap (`config-sigstore-keys`). The TrustRoot reconciler doesn't run automatically when upgrading the webhook, right? How would you ensure that the ConfigMap gets updated when they upgrade?
- To avoid breaking changes to the existing `v1alpha.SigstoreKeys` (k8s type), I was considering adding a new field to `TrustRootSpec` (e.g. `TrustedRootJSON:`) which would allow the user to just paste in a copy of a `trusted_root.json` (e.g. [this one from root-signing](https://github.com/sigstore/root-signing/blob/main/targets/trusted_root.json)). It seems simplest/safest to just accept it as a multiline string so that we don't need to maintain a yaml schema for it and the embedded `mediaType` could be used to deserialize it (within sigstore-go). Does this seem like a decent approach?

hectorj2f:
> have the same question about the validRepository/rootJSON values in the tests (serialized TUF repos). You don't currently have a script to do this, do you? I see you updated them somewhat recently. If you have any notes, that would be helpful -- I'm assuming you used the go-tuf CLI? Edit: Regarding the first two bullets, I finally read the comment stating that these values were pulled from a scaffolding deployment on a kind cluster! I did go ahead and move these values to separate files in a testdata subdir and added a minimal script to regenerate the marshalledEntry which saved me some time debugging since the diff output from TableTest is hard to read.

@codysoyland Would you mind creating a separate PR for this? So we can add these changes to main.

hectorj2f:
> The TrustRoot reconciler doesn't run automatically when upgrading the webhook, right? How would you ensure that the ConfigMap gets updated when they upgrade?

When you restart the webhook, it tries to reconcile all the TrustRoot resources.

If I understood you correctly. We have a watch on the configMap so every time it gets updated, we get the latest changes.

hectorj2f:
> For now, this PR will be constrained to only affecting internal data types and pulling the trusted root from TUF.

This sounds great. 
For any breaking change, we probably want to open one issue and discuss any decision with @vaikas and others.

codysoyland:
@hectorj2f I rebased this PR onto https://github.com/sigstore/policy-controller/pull/1325, added coverage for the fetching of `trusted_root.json` from TUF, and spent some time (along with @bdehamer) fixing all the tests/linter issues so I will mark it ready for review. After #1325 is merged, the diff will become a little smaller. Thanks in advance for your review!

.github/workflows/build.yaml
```diff
@@ -43,7 +43,7 @@ jobs:
           check-latest: true
 
       # will use the latest release available for ko
-      - uses: imjasonh/setup-ko@ace48d793556083a76f1e3e6068850c1f4a369aa # v0.6
+      - uses: ko-build/setup-ko@ace48d793556083a76f1e3e6068850c1f4a369aa # v0.6
```
hectorj2f, line 46: I'd suggest to make these changes as part of a different PR

Makefile
```diff
@@ -189,3 +189,6 @@ docs/generate-api:
 	    `find ./pkg/apis/policy/v1alpha1/ -iname '*types.go' |  sort -r | tr '\n' ' '` \
 	    > docs/api-types/index-v1alpha1.md;
 
+.PHONY: generate-testdata
+generate-testdata:
```
hectorj2f, line 193: It seems you need to rebase, I know we have merged this code.

vaikas (commented):
Nice! Not too bad after all, love it when a plan comes together 👍 Just need to do some rebasing as Hector says, but looks good overall! Thanks for handling my TODOs :)

.github/workflows/build.yaml
codysoyland: Oddly, this failed to build in our private fork as the setup-ko action has moved -- not sure why it works here! I'll move them to a different PR.

codysoyland: https://github.com/sigstore/policy-controller/pull/1328

pkg/apis/config/sigstore_keys.go
```diff
@@ -133,56 +87,171 @@ func NewSigstoreKeysFromConfigMap(config *corev1.ConfigMap) (*SigstoreKeysMap, e
...
+		case 3072:
+			keyDetails = pbcommon.PublicKeyDetails_PKIX_RSA_PSS_3072_SHA256
+		case 4096:
+			keyDetails = pbcommon.PublicKeyDetails_PKIX_RSA_PSS_4096_SHA256
+		default:
+			keyDetails = pbcommon.PublicKeyDetails_PUBLIC_KEY_DETAILS_UNSPECIFIED
+		}
+	default:
+		keyDetails = pbcommon.PublicKeyDetails_PUBLIC_KEY_DETAILS_UNSPECIFIED
+	}
```
codysoyland, line 246: I need to spend a few moments to confirm these assumptions with regard to key type. @kommendorkapten, if you have a moment to weigh in, that would be helpful! This code path is exercised when a transparency log public key is provided in a custom resource (PEM encoded), or when a TUF repo does not have a `trusted_root.json`. So it does not apply to the Sigstore Public Good instance, which provides a `trusted_root.json`. For private transparency logs, we need to assume the key type to fill in the values required by the TrustedRoot.

kommendorkapten: I think this code is solid. It uses Go std `ParsePKIXPublicKey`, and the created public key should so have the correct parameters.

The only thing that possibly could break is for RSA is the intended signature scheme is PKCS1#v1.5, but without any OOB information, we can't tell which signature scheme to use.

Do you know where in the code we load the verifier for RSA key? I can also take a look, but there are two options in sigstore/sigstore:
PKCS#1v1.5: https://github.com/sigstore/sigstore/blob/main/pkg/signature/rsapkcs1v15.go#L121
PSS: https://github.com/sigstore/sigstore/blob/main/pkg/signature/rsapss.go#L147

kommendorkapten: Ok, so I think we should change things here.

There are two places I found where a verifier is loaded 
* https://github.com/sigstore/policy-controller/blob/fc2800a5e9cf28f75c63e5ac6f50e2e468f6b5a2/pkg/webhook/validation.go#L38 
* https://github.com/sigstore/policy-controller/blob/fc2800a5e9cf28f75c63e5ac6f50e2e468f6b5a2/pkg/webhook/validator.go#L839

Both of them calls `sigstore/pkg/signature/verifier::LoadVerifier`. That functions will assume PKCS#1v1.5 padding. To use PSS padding, you need to explicitly load that via `signature.LoadRSAPSSVerifier`, there are no such references in this code base.

So in the code above, you should load:
* `PKIX_RSA_PKCS1V15_2048_SHA256`
* `PKIX_RSA_PKCS1V15_3072_SHA256`
* `PKIX_RSA_PKCS1V15_4096_SHA256`

IMHO this should be clarified in the docs too, that when using RSA keys, PKCS#1v1.5 padding is only supported.

pkg/apis/config/sigstore_keys.go
```diff
@@ -133,56 +87,171 @@ func NewSigstoreKeysFromConfigMap(config *corev1.ConfigMap) (*SigstoreKeysMap, e
...
+			keyDetails = pbcommon.PublicKeyDetails_PKIX_ECDSA_P384_SHA_384
+		case elliptic.P521():
+			keyDetails = pbcommon.PublicKeyDetails_PKIX_ECDSA_P521_SHA_512
+		default:
+			keyDetails = pbcommon.PublicKeyDetails_PUBLIC_KEY_DETAILS_UNSPECIFIED
+		}
+	case *rsa.PublicKey:
+		switch k.Size() {
+		case 2048:
+			keyDetails = pbcommon.PublicKeyDetails_PKIX_RSA_PSS_2048_SHA256
```
kommendorkapten, line 236:
```suggestion
			keyDetails = pbcommon.PublicKeyDetails_PKIX_RSA_PKCS1V15_2048_SHA256
```

pkg/apis/config/sigstore_keys.go
```diff
@@ -133,56 +87,171 @@ func NewSigstoreKeysFromConfigMap(config *corev1.ConfigMap) (*SigstoreKeysMap, e
...
+		case 3072:
+			keyDetails = pbcommon.PublicKeyDetails_PKIX_RSA_PSS_3072_SHA256
```
kommendorkapten, line 238:
```suggestion
			keyDetails = pbcommon.PublicKeyDetails_PKIX_RSA_PKCS1V15_3072_SHA256
```

pkg/apis/config/sigstore_keys.go
```diff
@@ -133,56 +87,171 @@ func NewSigstoreKeysFromConfigMap(config *corev1.ConfigMap) (*SigstoreKeysMap, e
...
+		case 4096:
+			keyDetails = pbcommon.PublicKeyDetails_PKIX_RSA_PSS_4096_SHA256
```
kommendorkapten, line 240:
```suggestion
			keyDetails = pbcommon.PublicKeyDetails_PKIX_RSA_PKCS1V15_4096_SHA256
```

pkg/apis/config/sigstore_keys.go
codysoyland: Thanks @kommendorkapten! I will file an issue for adding the ability use a protojson trusted root in the TrustRoot CRD which will make this section of code obsolete, and by the way, these key details are not actually used anywhere in Policy Controller at the moment -- They will only be used when we add the ability to use Sigstore Bundles as OCI artifacts. So this is not a regression and the trusted root support should be complete before we add any support for bundles.

codysoyland:
I rebased again and also removed unnecessary go.mod updates, which in turn removed the commits related to run-running codegen, so I think this PR is in a good state to merge now. Would love to have this reviewed again @hectorj2f @vaikas 🙏🏻  Thanks!

hack/gentestdata/gentestdata.go
```diff
@@ -1,10 +1,10 @@
-// Copyright 2024 The Sigstore Authors
+// Copyright 2022 The Sigstore Authors.
```
hectorj2f, line 1: 2024 ?

hectorj2f (dismissed):
LGTM overall

hectorj2f (approved):

pkg/apis/config/sigstore_keys.go
```diff
@@ -133,56 +87,171 @@ func NewSigstoreKeysFromConfigMap(config *corev1.ConfigMap) (*SigstoreKeysMap, e
...
 // TransparencyLogInstance.
-func ConvertTransparencyLogInstance(source v1alpha1.TransparencyLogInstance) TransparencyLogInstance {
-	return TransparencyLogInstance{
-		BaseURL:       *source.BaseURL.DeepCopy(),
-		HashAlgorithm: source.HashAlgorithm,
-		PublicKey:     source.PublicKey,
+func ConvertTransparencyLogInstance(source v1alpha1.TransparencyLogInstance) *pbtrustroot.TransparencyLogInstance {
+	pbpk, pk, err := DeserializePublicKey(source.PublicKey)
+	if err != nil {
+		return nil // TODO: log error? Add return error?
```
vaikas, line 147: Yeah, I think we should return the error instead of silently eating it. Here and in L151, since we return it from DeserializePublicKey below.

hectorj2f (approved):

vaikas (approved):
Thanks!

PR #1458, opened 2024-06-03
Add resync period functionality for TrustRoot resource

malancas:
#### Summary

As I understand it, the TrustRoot resource is only created on startup. This means that if the user specifies a remote trust root source in a policy, the TUF cache will only be created once and is never updated. This PR adds a resync period for the trust root resource, so if a remote trust root source is specified, the TUF cache should be updated when the resync happens. I set the default resync period for 24 hours. I followed the [policy resync functionality](https://github.com/sigstore/policy-controller/blob/main/pkg/reconciler/clusterimagepolicy/controller.go#L110) as a guide here. It made sense to follow the same pattern.


vaikas:
Right on!

cmd/webhook/main.go
```diff
@@ -98,6 +98,10 @@ var (
 	// policyResyncPeriod holds the interval which ClusterImagePolicies will resync
 	// This is essential for triggering a reconcile update for potentially stale KMS authorities.
 	policyResyncPeriod = flag.String("policy-resync-period", "10h", "The resync period for ClusterImagePolicies. The default is 10h.")
+
+	// trustrootResyncPeriod holds the interval which the TrustRoot will resync
+	// This is essential for triggering a reconcile update for potentially stale TUF metadata.
+	trustrootResyncPeriod = flag.String("trustroot-resync-period", "24h", "The resync period for ClusterImagePolicies. The default is 24h.")
```
codysoyland, line 104: Can you use [flag.Duration](https://pkg.go.dev/flag#Duration) here?

.github/workflows/kind-cluster-image-trustroot-resync-period.yaml
codysoyland, line 1: Does this e2e test only check the flag validation behavior? IMHO, standing up an e2e test just for this is heavy-handed.

My suggestion is either to remove it (I'm not aware that we test other flag validation), or convert the existing flags to a [Flagset](https://pkg.go.dev/flag#FlagSet.Parse) and write a unit test for flag validation.

cmd/webhook/main.go
malancas: Yes, looks like we can. I'll update this.

.github/workflows/kind-cluster-image-trustroot-resync-period.yaml
malancas: That's correct. There was already an e2e validating the policy resync period flag behavior that I copied for this one. But since we can use `flag.Duration` for both the `trustroot-resync-period` and `policy-resync-period` flags, I think we can just this e2e test and the policy resync one altogether.

hectorj2f (dismissed):

malancas:
@hectorj2f @vaikas now that the `policy-resync-period` flag is handled with `flag.Duration`, I dropped the `.github/workflows/kind-cluster-image-policy-resync-period.yaml` test since the custom flag parsing and logging is no longer needed. But let me know if we still want this test otherwise.

codysoyland (approved):

hectorj2f (approved):

Issue #1388, opened 2024-04-29
Verify only pods


dszakallas:
**Description**

Make policy-controller configurable to only consider pods, not higher level resources. 

### Use case

We are using a continuous reconciliation (GitOps) solution, Flux, to maintain cluster state. The mutation of the image reference  by the mutating admission webhook has indesired effects for us when used in conjunction with Flux:
- Jobs are immutable. It can happen that digest changes for the same tag between reconciliations, causing an mutation to happen to a job, which is rejected by kube api server.
- even for deployments, resolving a digest every sync cycle causes a tight control that forces redeployment whenever the digest changes behind a tag, which we find undesirable.

### Workarounds

While the scope of resources to validate can be controlled on the Policy level, this is not sufficient for our use, as mutations still get applied to all recognized resources and this [seems to be hard coded](https://github.com/sigstore/policy-controller/blob/main/cmd/webhook/main.go#L202-L209). Eventually, the only workaround I found was to fork the code and remove these, alond with their corresponding resource names.

### Relwork

Correctly me if I am wrong, but I believe this change would not deteriorate the integrity of the cluster, since the pods owned by these unvalidated resources would still go through admission control, and their images will be verified. As an example of this practice, PSA (Pod Security Admission) only operates on the pod resource. Recently, Connaisseur also [added a flag](https://github.com/sse-secure-systems/connaisseur/releases/tag/v3.4.0) to opt out of higher level resource validation.

senanz:
@hectorj2f - could you please assign it to me, started to work on code that supporting that.

0xiso:
@senanz We are also facing the same issue and are searching for solutions. Seems like you are working on solving this problem, but do you have plans to create a PR?

hectorj2f:
@0xiso @senanz feel free to assign it yourself, but we need to discuss whether we want to go ahead with this change.

senanz:
@hectorj2f - I have already have the change, need to do some modification and let's discuss that then.
will update you soon.
Thanks for consering that.

senanz:
@0xiso - I have the PR need some updates and will raise it again.
@hectorj2f -Could you please assign the ticket on me?

senanz:
@hectorj2f - Could you please review?
https://github.com/sigstore/policy-controller/pull/1687
Let me know of you want to do dsicussion regarding that on slack.

Cajga:
[...]

PR #1629, opened 2024-09-25
https://github.com/sigstore/policy-controller/issues/1388, Support co…

senanz:
…ntrol what resource to be monitored from the chart instead hard coded in the code


hectorj2f (changes requested):
I don't understand the benefits of this PR.

senanz:
Hi @hectorj2f , Today all resources that will be monitored by the policy controller is hardcoded, we want to control that from chart values, For example if i want to limit the policy controller to monitor just the pods.

Mukuls77:
@hectorj2f  i wanted to share my views on this PR. This PR is not really removing the functionality of modifying the Image Ref in Pod sepc from IMG:TAG to IMG@Digest but instead of making it a hardcode logic this PR bring a feature flag --disable-mutation which if used by the user will disable the required mutation. also the PR want to introduce configurability on which k8s resources should be monitored for Image verification so user can choose to include replicaset or not for image verification based on his/her need.
This way policy controller will be more configurable and usable.
In the current state with mandatory TAG to Digest conversion we see issues in introduction of Policy controllers in the setups where there are already some existing implementations which expect TAG in podspec to track the pods, introduction of policy controller will thus break the existing Application logic.
more over mandatory checking of Image signature for replicaset, daemonset with Mutating TAG to Digest introduces a big issue as now we can not introduce policy controller in a setup to do a Inservice upgrade or let say scale-in scale-out a resource after introducing policy controller , as after policy controller introduction the podspec get mutated to digest which changes the resource creation , k8s controller detect this change and keep on creating the given resource.
so in my view these configuration bring more usability to policy controller. 
If there is some common call scheduled we would like to share our views on this PR.

Issue #1657, opened 2024-10-09
ACR credential helper blocks when using non-ACR registries


codysoyland:
**Description**

The ACR credential helper seems to block execution when using non-ACR images in some clusters. I've observed this behavior in GKE, but not in local kind clusters.

The blocking call occurs on [this line in the ACR credential helper](https://github.com/sigstore/policy-controller/blob/25afffa1ffe3bbcb5dcefb7e0cabbdc712c9cd09/pkg/webhook/registryauth/azure/acrhelper.go#L55).

This results in pods failing admission due to the webhook timeout elapsing:

```
error: failed to create deployment: Internal error occurred: failed calling webhook "policy.sigstore.dev": failed to call webhook: Post "https://webhook.cosign-system.svc:443/mutations?timeout=25s": context deadline exceeded
```

The commit that introduced this bug is f099546d3ce22170217ac170191c3b7e97371ab9, which is between Policy Controller v0.9.0 and v0.10.0 (neither version is yet referenced by the [Helm chart](https://github.com/sigstore/helm-charts/blob/main/charts/policy-controller/Chart.yaml), hence why this bug has seen little visibility). It seems the upstream `azidentity` package made a change that causes the call to `DefaultAzureCredential.GetToken` to block. I believe this has something to do with timeout modifications that were made in that version of the `azidentity` library.

I have a fix I will publish shortly that will only attempt to use this helper if the registry is a subdomain of `azurecr.io`.

**Version**

v0.10.0

codysoyland:
This seems to be fixed in https://github.com/sigstore/policy-controller/pull/1656. My PR https://github.com/sigstore/policy-controller/pull/1658 is still a good practice though, in my opinion.

PR #1658, opened 2024-10-09
Add check for ACR registry in ACR credential helper

codysoyland:
#### Summary

This PR adds a check that will only run the ACR credential helper when the target registry is a subdomain of `azurecr.io`.

This is similar to an [existing check used for GCR](https://github.com/google/go-containerregistry/blob/main/pkg/v1/google/keychain.go#L62-L64).

Fixes https://github.com/sigstore/policy-controller/issues/1657

### DIFF
--- a/pkg/webhook/registryauth/azure/acrhelper.go
+++ b/pkg/webhook/registryauth/azure/acrhelper.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"net/url"
 	"os"
+	"strings"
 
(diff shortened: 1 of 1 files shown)


malancas (approved):

codysoyland:
The original issue seems to be resolved by https://github.com/sigstore/policy-controller/pull/1656, which merged earlier today.

I still think this PR is good to merge, as it will speed things along slightly if ACR is not used, but the fix has basically been implemented.

hectorj2f (approved):

hectorj2f:
Thanks @codysoyland although there is one CI check failing.

hectorj2f (approved):

PR #1687, opened 2024-11-03
Control the policy controller monitoring resources from chart, enhanc…

### LINKED ISSUE #1388: Verify only pods

senanz:
…ment for the current implmentation to hadd all the resources by default, The new change add avail to pass resourcesNames through the chart with list of resources comma sperataed for which resources to be monitored by the policy controller, the default is all resources if the flag wasn't presented in the chart

This PR resolves https://github.com/sigstore/policy-controller/issues/1388 

Signed-off-by: Senan Zedan (EXT-Nokia) <senan.zedan.ext@nokia.com>

#### Summary
 The new change add avail to pass resourcesNames through the chart with list of resources comma sperataed for which resources to be monitored by the policy controller, the default is all resources if the flag wasn't presented in the chart


senanz:
@vaikas - Could you please look into this?

vaikas:
Could you add some tests, maybe here: https://github.com/sigstore/policy-controller/tree/main/test

vaikas:
> @vaikas - per your request, test added.

That does not really test any of this new code. You'd have to add a test that launches policy controller with these new flags, here's one example where we change the policy-controller behaviour by the flags:
https://github.com/sigstore/policy-controller/blob/b3af2f053ba9b652c96a99b08580e75ed2788cbc/.github/workflows/kind-cluster-image-policy-tsa.yaml#L125

So, create a new kustomize file that launches policy controller with say --resource-name=pods, and customize it like here:
https://github.com/sigstore/policy-controller/blob/b3af2f053ba9b652c96a99b08580e75ed2788cbc/.github/workflows/kind-cluster-image-policy-tsa.yaml#L136

And then after the policy-controller has been started with the flags under test, you'd run your new test:
https://github.com/sigstore/policy-controller/blob/b3af2f053ba9b652c96a99b08580e75ed2788cbc/.github/workflows/kind-cluster-image-policy-tsa.yaml#L177

It should at the bare minimum have a negative test and a positive test. So, if you're using resource-name pods, then launching a deployment with a failing config should succeed if you instead launch a pod.

.github/workflows/kind-cluster-custom-resources.yaml
```diff
@@ -0,0 +1,180 @@
...
+
+defaults:
+  run:
+    shell: bash
+
+permissions: read-all
+
+jobs:
+  cip-test-trustroot-bring-your-own-keys:
+    name: ClusterImagePolicy e2e tests with TrustRoot - Bring Your Own Keys
```
vaikas, line 29: Nit: Change the test name, and the job name to match your new functionality.

.github/workflows/kind-cluster-custom-resources.yaml
```diff
@@ -0,0 +1,180 @@
...
+        uses: chainguard-dev/actions/setup-mirror@main
+        with:
+          mirror: mirror.gcr.io
+
+      - uses: sigstore/cosign-installer@dc72c7d5c4d10cd6bcb8cf6e3fd625a9e5e537da # v2
+
+      - name: Install cluster + sigstore
+        uses: sigstore/scaffolding/actions/setup@main
+        env:
+          INSTALL_TSA: true
```
vaikas, line 120: I don't think you necessarily need this, but 🤷

.github/workflows/kind-cluster-custom-resources.yaml
```diff
@@ -0,0 +1,180 @@
...
+          POLICY_CONTROLLER_ARCHS: linux/amd64
+        run: |
+          make ko-policy-controller
+          kustomize build test/kustomize-custom-resource | kubectl apply -f -
+
+          # Wait for the webhook to come up and become Ready
+          kubectl rollout status --timeout 5m --namespace cosign-system deployments/webhook
+          echo "TUF_ROOT_FILE=./root.json" >> $GITHUB_ENV
+
+      - name: Checkout TSA for testing.
```
vaikas, line 141: And you probably don't need this either here since we don't test TSA.

.github/workflows/kind-cluster-custom-resources.yaml
```diff
@@ -0,0 +1,180 @@
...
+          if ! ./timestamp-cli verify --timestamp response.tsr --artifact "myblob" --certificate-chain ts_chain.pem ; then
+            echo "failed to verify timestamp"
+            exit -1
+          fi
+          if ! ./timestamp-cli inspect --timestamp response.tsr --format json ; then
+            echo "failed to inspect the timestamp"
+            exit -1
+          fi
+
+      - name: Run Custom resources Tests
```
vaikas, line 173: You can remove L141 to here, since this is the only thing we're really testing here. No need to test the TSA again.

test/e2e_test_policy_custom_resource.sh
```diff
@@ -0,0 +1,68 @@
...
+# Ensure a non-monitored resource is NOT included
+if kubectl get MutatingWebhookConfiguration -o yaml | grep -q "resources:.*DaemonSet"; then
+  echo "DaemonSet should not be included in MutatingWebhookConfiguration"
+  exit 1
+else
+  echo "DaemonSet correctly excluded from MutatingWebhookConfiguration"
+fi
+echo '::endgroup::'
+
+echo '::group:: Cleanup'
```
vaikas, line 64: I'd add two more test cases, something like:
`kubectl create deployment ...` with container image that has a tag
Then `kubectl get deployment` and check that the container image in the spec has not been modified, and then `kubectl get pod` and check that the container image in the spec has been modified.

test/testdata/policy-controller/e2e/test-deployment-with-custom-resource.yaml
```diff
@@ -0,0 +1,36 @@
...
+  template:
+    metadata:
+      labels:
+        app: demo-custom-resource
+    spec:
+      containers:
+          - args:
+              - resource-name="pods"
+          - name: demo-custom-resource
+          image: TEST_IMAGE
```
vaikas, line 36: The problem here is that the TEST_IMAGE as created has a full digest in it so it's harder to tell if it's been resolved to digest or not. I'd suggest that you have an image that has a tag, which would without this change be resolved to digest for a deployment, so creating a deployment with a tag here would be the simplest way to verify the functionality works. Then the resulting pods should have the digests resolved.

test/testdata/policy-controller/e2e/test-deployment-with-custom-resource.yaml
```diff
@@ -0,0 +1,36 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  namespace: demo-scalable-custom-resource
```
vaikas, line 18: nit: use a different namespace, say: test-deployment-custom-resource

vaikas (changes requested):
Looking great! There's just some nits for naming, namespaces. One change that should be made it so make sure the tag->digest resolution does not actually happen. The webhook configuration is a good check, but it's easy enough to also make sure the actual resource (deployment) does not get modified.

.github/workflows/kind-cluster-custom-resources.yaml
senanz: Acknolwedged, let's have it now for the simplicty.

.github/workflows/kind-cluster-custom-resources.yaml
senanz: You are right, removed.

.github/workflows/kind-cluster-custom-resources.yaml
senanz: Removed.

test/testdata/policy-controller/e2e/test-deployment-with-custom-resource.yaml
senanz: Thanks, updated.

vaikas:
Looks like some of the changes I requested, and you marked as fixed have not made it in to this PR 🤔 ?

Also, looks like bunch of `sed` commands are failing, for example:
https://github.com/sigstore/policy-controller/actions/runs/12234409121/job/34140387122?pr=1687#step:14:132

senanz:
Hi Ville,
Fixed them locally and working in the last comment to add tests.
Updated PR will raise soon.

cmd/webhook/main.go
```diff
@@ -78,6 +79,13 @@ var (
 	// https://github.com/sigstore/policy-controller/issues/354
 	disableTUF = flag.Bool("disable-tuf", false, "Disable TUF support.")
 
+	// Validate specific resources.
+    // https://github.com/sigstore/policy-controller/issues/1388
```
vaikas, line 83: nit, but the format looks off here?

cmd/webhook/main.go
```diff
@@ -116,6 +124,8 @@ func main() {
 	flag.IntVar(&opts.Port, "secure-port", opts.Port, "The port on which to serve HTTPS.")
 
 	flag.Parse()
+	resourcesNamesList := strings.Split(*resourcesNames, ",")
+    listResources = append(listResources, resourcesNamesList...)
```
vaikas, line 128: and here. Maybe run gofmt on these files?

.github/workflows/kind-cluster-custom-resources.yaml
```diff
@@ -0,0 +1,148 @@
...
+jobs:
+  cip-test-trustroot-bring-your-own-keys:
+    name: ClusterImagePolicy e2e tests with custom resources
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false # Keep running if one leg fails.
+      matrix:
+        k8s-version:
+          - v1.27.x
```
vaikas, line 36: @cpanato is updating the tested k8s versions, should proabably bump these once it merges.

cmd/webhook/main.go
senanz: seems something related to git.
I'm using intelji and format looks good.
image

cmd/webhook/main.go
senanz: Same as the above comment the format at the editor looks fine.

.github/workflows/kind-cluster-custom-resources.yaml
senanz: @cpanato - Could you please let me know when you merge your changes so i can make the right adjcments here?

cmd/webhook/main.go
vaikas: The editor might display things with spaces vs. tabs, but gofmt which ensures that regardless of which ide you use the code looks the same is not agreeing, for example:
https://github.com/sigstore/policy-controller/actions/runs/12412408356/job/34695099927?pr=1687

Did you try running gofmt? At the root if you do:
```
gofmt -s -w .
```

And then it should update everything for you.

test/testdata/policy-controller/e2e/test-deployment-with-custom-resource.yaml
senanz: Test was added.

cmd/webhook/main.go
senanz: Ran the command: gofmt -s -w .

cmd/webhook/main.go
senanz: Done

.github/workflows/kind-cluster-custom-resources.yaml
vaikas: This was done, so go ahead and adjust this matrix.

senanz: Updated to:
- v1.29.x
- v1.30.x
- v1.31.x
- v1.32.x

.github/workflows/kind-cluster-custom-resources.yaml
```diff
@@ -141,6 +141,7 @@ jobs:
       - name: Run Custom resources Tests
         timeout-minutes: 5
         run: |
+          chmod +x ./test/e2e_test_policy_custom_resource.sh
```
vaikas, line 144: FYI, you should be able to make this change in git so that you don't have to do this here.

senanz:
hi @0xiso , Will back to work on that early next week, will ping you on slack and for sure we can collaborate.

Cajga:
Hi @senanz , sorry for bothering but do you still plan to give this a go?

test/e2e_test_policy_custom_resource.sh
```diff
@@ -0,0 +1,114 @@
+#!/usr/bin/env bash
+#
+# Copyright 2024 The Sigstore Authors.
```
hectorj2f, line 3: 2025

Issue #1704, opened 2024-11-15
Add support for TrustedRoot in air-gapped environments


codysoyland:
**Description**

I would like to support the [TrustedRoot JSON format](https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_trustroot.proto) in a simple way for air-gapped environments.

Currently, there are 3 ways to specify the trusted keys in the TrustRoot CRD:

- Remote: Using a root JSON and TUF URL, this will fetch the trusted root via TUF
- Repository: Using a root JSON and serialized TUF filesystem (base64, tarred, gzipped set of files, aka MirrorFS), this allows for air-gapped mode starting from a TUF repository
- SigstoreKeys: This struct provides much of the same functionality of the TrustedRoot JSON, but is missing validity time periods, and requires conversion to the TrustedRoot format as many of the data types are subtly different (for example, certs are represented as DER in TrustedRoot, and PEM in SigstoreKeys)

My proposal is to add a `TrustedRootJSON` type (either string or []byte) to the CRD that contains the trusted root JSON as defined by the protobuf spec. This would allow for full support of trusted root in the simplest way possible.

PR #1705, opened 2024-11-15
Add TrustedRootJSON to TrustRootSpec

codysoyland:
This adds `TrustedRootJSON` to the `TrustRootSpec` to allow using of the [`TrustedRootJSON` spec](https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_trustroot.proto) in air-gapped environments.

usage:
```yaml
kind: TrustRoot
metadata:
  name: my-trusted-root
spec:
  trustedRootJSON: |
    {
      "mediaType": "application/vnd.dev.sigstore.trustedroot+json;version=0.1",
      "certificateAuthorities": [
        {
...
```

Note this uses type `string`. We could alternatively use type `[]byte` for this, which would mean that the `trustedRootJSON` value would need to be base64-encoded. That would be perhaps less prone to human-error in indentation. I'm happy to make that change if deemed better by the maintainers.

Fixes https://github.com/sigstore/policy-controller/issues/1704

### DIFF
--- a/docs/api-types/index-v1alpha1.md
+++ b/docs/api-types/index-v1alpha1.md
@@ -138,6 +138,7 @@ TrustRootSpec defines a trusted Root. This is typically either a TUF Root or a b
 | remote | Remote specifies initial root of trust & remote mirror. | [Remote](#remote) | false |
 | repository | Repository contains the serialized TUF remote repository. | [Repository](#repository) | false |
 | sigstoreKeys | SigstoreKeys contains the serialized keys. | [SigstoreKeys](#sigstorekeys) | false |
+| trustedRootJSON | TrustedRootJSON contains the serialized trusted_root.json. | string | false |
 
 [Back to TOC](#table-of-contents)
 


federico-falconieri-form3:
@malancas @hectorj2f @vaikas what do you think of this one?

vaikas:
Regarding the format, I think you can create the json as a string as a user and not worry about the indentation , so not sure if it's worth doing the roundtrip to/from base64, but I don't really feel strongly about that. The validation here should catch any indentation errors:
https://github.com/sigstore/policy-controller/pull/1705/files#diff-b3bd6cb5089d48b1e5b33b8a00ea3b4239e42b17d34ff6cbc92227caa7c272a6R49

Can you add a test for this, something like:
https://github.com/sigstore/policy-controller/blob/main/test/e2e_test_cluster_image_policy_with_trustroot_repository.sh

Another way to do this might be to hoist the api definition to v1beta1 for trustroots, and at that point change the previous field(s) to not support the old version, and only support the TrustedRootJSON.

As in, the trust root is still in v1alpha1, and I'm curious if that should be hoisted to v1beta1, and get a deprecation plan for getting rid of v1alpha1 (obviously not part of this PR, but just thinking out loud).

codysoyland:
> Can you add a test for this, something like:
https://github.com/sigstore/policy-controller/blob/main/test/e2e_test_cluster_image_policy_with_trustroot_repository.sh

👍🏻 

> As in, the trust root is still in v1alpha1, and I'm curious if that should be hoisted to v1beta1, and get a deprecation plan for getting rid of v1alpha1 (obviously not part of this PR, but just thinking out loud).

No strong feelings here... perhaps the deprecation plan would be to support both in v1alpha1 and only trustroot json in v1beta1? I'm okay with either option!

PR #1747, opened 2024-12-20
fix post submit job

cpanato:
- fix post submit job

see: https://github.com/sigstore/policy-controller/actions/runs/12433003788/job/34713622190


vaikas (approved):
Thanks!

PR #1751, opened 2025-01-03
Use v0.7.18 for scaffolding, update k8s versions to test with.

vaikas:
#### Summary
Remove old versions of k8s since they are no longer supported, we should not be testing with them.
Use new scaffolding version that support later versions of k8s.


hectorj2f (approved):

cpanato (approved):
thanks

PR #1753, opened 2025-01-03
fix goreleaser signning part

cpanato:
see https://github.com/sigstore/policy-controller/actions/runs/12599069304/job/35116178539

to fix that we set the paralellism to 1, we did this in other pipelines


vaikas (approved):
Thanks!

Issue #810, opened 2023-05-31
Feature Request: namespaced ImagePolicy


michaelst:
It would be nice to have a namespaced resource (instead of entire cluster) for image policies. For example I want to say a images can only be verified if in the a namespace, and b images only in the b namespace.

Horiodino:
Hey, if no one is working on this, I would like to take it on.

PR #1780, opened 2025-02-02
Namespaced image policy

Horiodino:
#### Summary

CIP is verified against namespace policies. If deployed in the correct namespace, validation passes; otherwise, an error is returned. Namespaces not defined in policies are ignored.

fixes: https://github.com/sigstore/policy-controller/issues/810

### DIFF
--- a/pkg/apis/policy/v1alpha1/clusterimagepolicy_conversion.go
+++ b/pkg/apis/policy/v1alpha1/clusterimagepolicy_conversion.go
@@ -175,6 +175,9 @@ func (p *Policy) ConvertTo(_ context.Context, sink *v1beta1.Policy) {
 	if p.IncludeTypeMeta != nil {
 		sink.IncludeTypeMeta = ptr.Bool(*p.IncludeTypeMeta)
 	}
+	if p.NamespaceSelector != "" {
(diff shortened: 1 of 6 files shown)


Horiodino:
can anyone review this please .

hectorj2f:
@Horiodino We should discuss on the issue whether this is intended behaviour or not.

Horiodino:
😅 I noticed the issue was open, so I went ahead and worked on it. I did ask before starting but didn’t receive a response. I’m here for any discussion if there’s a need to revisit this.

michaelst:
I think this would be a great feature to support. My specific need is a I have multiple different teams deploying to the same cluster, it would be great to make sure images sourced from a specific GitHub repository are only deployable to the assigned namespace.

Horiodino:
> I think this would be a great feature to support. My specific need is a I have multiple different teams deploying to the same cluster, it would be great to make sure images sourced from a specific GitHub repository are only deployable to the assigned namespace.

Yeah, I agree. It makes sense since certain teams often have access to specific namespaces, and having these restrictions in place adds an extra layer of control.

0xiso:
I also strongly agree with the addition of this feature. It is a necessary feature for clusters used by multiple teams.

Issue #1681, opened 2024-10-27
How to stop the Image tag of managed sidecar from mutating?


sjain700:
**Description**

This is a follow-up for the following question.
https://github.com/sigstore/policy-controller/issues/1220

I can see that sidecar tags is mutated and the version is gone. Can there be a enhancement where the version is kept and sha is appended. Something like **image:version@SHA.**

PR #1824, opened 2025-04-16
Preserve original tag when resolving an image tag to digest

0xiso:
#### Summary

This PR updates the code to preserve the original image tag when resolving image tags to digests in the mutating webhook.

Currently, the Sigstore Policy Controller's mutating webhook converts image references from `{registry}/{repository}:{tag}` to `{registry}/{repository}@sha256:{hash}`. This PR changes the behavior to convert them to `{registry}/{repository}:{tag}@sha256:{hash}` instead.

This change is considered to have no security concerns because when both a tag and a digest are specified for an image, the tag is ignored, and only the digest is used for image pulling and verification by container runtimes.

Preserving the image tag offers benefits such as:

* **Improved human readability:** Makes it easier to quickly identify which image version is being used.
* **Enhanced observability:** Useful for collecting service metrics based on image tags.

Closes https://github.com/sigstore/policy-controller/issues/1681

### DIFF
--- a/pkg/webhook/validator.go
+++ b/pkg/webhook/validator.go
@@ -1078,7 +1078,12 @@ func (v *Validator) resolvePodSpec(ctx context.Context, ps *corev1.PodSpec, opt 
 					logging.FromContext(ctx).Debugf("Unable to resolve digest %q: %v", ref.String(), err)
 					continue
 				}
-				cs[i].Image = digest.String()
+				// Keep the original tag and append the digest
+				if tagRef, ok := ref.(name.Tag); ok {
+					cs[i].Image = fmt.Sprintf("%s@%s", tagRef.Name(), digest.DigestStr())
+				} else {
+					cs[i].Image = digest.String()
+				}
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


0xiso:
@codysoyland Thank you for the comment. Indeed, _in the tests_, [this code](https://github.com/sigstore/policy-controller/blob/e58d7f5acbe8b8cfded12b04c99d72192d9503a9/pkg/webhook/validator_test.go#L1019-L1021) resolves the image reference to include both its tag and digest. However, the actual behavior is different; the image reference will resolve to include only the digest. I have [updated the test](https://github.com/sigstore/policy-controller/pull/1824/files#diff-9cb6e1e24474d4b69bc22d197f7e1817e74b376e4eb9d8975bb17df8f50f8440R1021) to reflect this. I also added a test case to cover where the original image is specified by digest only.

(I reverted the initial test update I made as it turned out to be unnecessary.)

codysoyland (approved):
Thank you for updating the test!

0xiso:
Hi @codysoyland, when is the code including this PR expected to be released?
We are waiting for the release of this code as the removal of an image tag is causing issues with our Datadog.

Issue #745, opened 2023-04-21
Disable update of "top level" objects


mlbiam:
**Question**
Is there a way to disable the policy-controller's rewriting of top level objects' `image` tags to specific digests?  For instance if I have the `image: docker.io/mlbiam/openuison-operator:0.0.2` in a `CronJob`, the `Pod` gets `image: index.docker.io/mlbiam/openunison-kubernetes-operator@sha256:a4af44351bfcf7fe1cfd05a9711f560127f6522508f89fd797e64d8a296d2159`, which makes sense.  The `CronJob` also gets this image though.  Is there a way to disable this behavior?

While I understand that goal of pinning the specific hash, it means that my GitOps controller is no longer the source of truth for my configuration (and could create a "thrashing" scenario where my controller and what is in cluster are constantly out of sync).

hectorj2f:
No. There isn't a way to disable the mutation from a tag to a digest. Using tags is a bad security practice. Your gitOps controller could use a digest with the image version, e.g. ldocker.io/library/golang:1.17.1@sha256:232a180dbcbcfa7250917507f3827d88a9ae89bb1cdd8fe3ac4db7b764ebb25a

mlbiam:
> Using tags is a bad security practice

digests are important because without a signature it's the only way for kubernetes to guarantee that the image being used is trusted.  With signature validation it's redundant.  I know the image is trusted because it's been validated against the identity used to sign it from the transparency log.  If I don't trust that identity, i don't trust the automated process that generates the digest that's put in the manifest.  it's redundant.

I get being "secure" by default, but this sort of rigidity is likely to lead to policy enforcement being turned off or being ignored.  It's not a great idea to make assumptions for cluster operators without context and without the option to let more advanced users configure exceptions based on their level of risk tolerance.  it will lead to a better UX and a more secure environment.

CodeGlitcher:
Kinda have the same problem.
We use a few kuberntes operators en for some it is not really easy to override every image to the digest. 

if the policy controller would have a option:
Juist validate the current digest of the selected tag on a pod start it would be perfect.
Our goal is juist to make sure that the pod we start is the pod we signed.

PR #1885, opened 2025-09-17
Introduce command line flag called pod-only

Cajga:
This PR introduces a command line flag called pod-only with a default value false. If set to true, only register the core/v1 Pod resource with the validating and mutating webhooks.

Resolves: #1388, #745, #285

If the code looks OK, I will look into creating test case(s).

#### Summary

Using the new flag will configure policy-controller to only mutate and validate `pod` resources. This will allow to use GitOps tools (like flux, argocd) to deploy workloads with version based tags without entering into a loop (gitops deployment with version tag -> policy controller mutates deployment with digest -> drift detected in tag so redeploy with version tag -> etc.). 

#### Release Note

* Introduced new command line flag called `pod-only`

#### Documentation

TBD


Cajga:
In fact, this is not needed as GitOps tools can use server side apply also for diff-ing/dry-runing and the mutating happens for that. See example here: https://github.com/sigstore/policy-controller/issues/1388#issuecomment-3309181944

PR #1894, opened 2025-10-13
Add OCI 1.1 referrers API support for attestation discovery

### LINKED ISSUE #1895: Policy controller cannot work with OCI 1.1 stored attestations
Google cloud build attestations are stored using the [OCI 1.1 spec](https://opencontainers.org/posts/blog/2024-03-13-image-and-distribution-1-1/), which introduces the referrer API as the new canonical way to link related artifacts such as images and their provenance attestations. As of now, the policy controller is unable to discover attestations stored this way, it can only discover attestations stored with the tag `<subject digest>.att`. If attestations exist but are stored with the OCI 1.1 method, then policy controller gives up with a `no matching attestations`.

In the example below I demonstrate how to reproduce this problem with the public image `us-docker.pkg.dev/cloudrun/container/hello@sha256:ee5d023051b0f808d0acb7880dec05ee0fc4f59667a529218ecf8159f1e72a45`. The 2 policies below are available [here](https://gist.github.com/falcorocks/efcb200a298599f97638a68bb696c7c9).

Image

We only need to instruct the controller to discover attestations using the OCI 1.1 spec, for instance if a flag is passed. I've implemented this in https://github.com/sigstore/policy-controller/pull/1894

falcorocks:
Closes #1895 

#### Summary

Some provenance attestation implementation, for instance Google Cloud Build, store attestations using the [OCI 1.1](https://opencontainers.org/posts/blog/2024-03-13-image-and-distribution-1-1/) referrers API rather than the traditional `<subject digest>.att` naming scheme. As is, the policy controller is thus unable to verify valid in-toto provenance attestations generated by Google Cloud Build. This is demonstrated in #1895.

This PR Implements OCI 1.1 referrers API for discovering attestations with automatic fallback to legacy tag-based discovery. This enables policy-controller to work with modern OCI 1.1 registries while maintaining backward compatibility, regardless if the new sigstore bundle format is used or not.

Technical implementation:
- Uses ociremote.Referrers() to discover attestation artifacts
- Filters by artifact types: in-toto, slsa, provenance, attestation, dsse
- Processes DSSE envelopes and creates compatible oci.Signature objects
- Maintains proper payload format for AttestationToPayloadJSON compatibility

### BEFORE

Image

### AFTER

Screenshot 2025-10-13 at 16 52 14

#### Release Note

- Add 'enable-oci11' configuration option
- Implement custom OCI 1.1 referrers API discovery in validAttestations()
- Add CLI flag '--enable-oci11' to policy-tester for testing
- Automatic fallback to legacy cosign discovery when OCI 1.1 fails

### DIFF
--- a/.github/workflows/kind-cluster-image-policy.yaml
+++ b/.github/workflows/kind-cluster-image-policy.yaml
@@ -41,6 +41,7 @@ jobs:
         script:
         - cluster_image_policy
         - cluster_image_policy_with_attestations
+        - cluster_image_policy_with_oci11_attestations
         - cluster_with_scalable
         - cluster_image_policy_with_warn
         - cluster_image_policy_with_source
--- a/cmd/tester/main.go
+++ b/cmd/tester/main.go
@@ -35,6 +35,7 @@ import (
 
 	"github.com/sigstore/policy-controller/pkg/apis/config"
 	"github.com/sigstore/policy-controller/pkg/apis/policy/v1alpha1"
+	policycontrollerconfig "github.com/sigstore/policy-controller/pkg/config"
 	"github.com/sigstore/policy-controller/pkg/policy"
 	"github.com/sigstore/policy-controller/pkg/webhook"
 )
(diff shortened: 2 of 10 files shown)


hectorj2f (approved):
lgtm

PR #1968, opened 2026-05-08
fix: Support cosign v3 signature verification

JasonPowr:
#### Summary
- Fix LogID encoding to store raw bytes instead of hex-encoded ASCII. cosign.GetTransparencyLogID() returns a hex string  the old code stored it via []byte(logID) (ASCII bytes of the hex chars) instead of hex.DecodeString(logID) (actual binary). This caused key-lookup mismatches with sigstore-go / cosign v3.

- Fix transparency log public key map lookups to use hex.EncodeToString() instead of string() on the now-correct raw byte LogIDs.
 
- Add bundle-format verification path for keyless and RFC3161Timestamp authorities, routing to attestation verification when NewBundleFormat is set.

#### Release Note
- Fixed LogID encoding in TrustRoot reconciliation to store raw bytes instead of hex-encoded ASCII, aligning with cosign v3 and sigstore-go expectations. This resolves transparency log key-lookup failures when verifying signatures against custom TrustRoot configurations.
- Added support for cosign v3 bundle-format signature verification for keyless and RFC3161Timestamp authorities.

#### Documentation
Users verifying cosign v3 bundle-format signatures can now set `signatureFormat: "bundle"` on a keyless authority in their ClusterImagePolicy. No changes are required for existing policies using the legacy signature format.

Custom TrustRoot configurations with transparency log entries (Rekor/CTFE) will now produce correctly encoded LogIDs. Existing TrustRoot resources should be re-reconciled after upgrading to ensure LogIDs are stored in the correct format.


JasonPowr:
@hectorj2f Sorry for the direct ping :), I seen you reviewed and approved the last pr that was merged into this repo, when you have the time, Could I get a review please :), thanks in advanced

JasonPowr:
@codysoyland Thank you so much for the review on my other [pr](https://github.com/sigstore/policy-controller/pull/1976) !, when you have a minute could I also get a review on this one ?

Issue #1974, opened 2026-06-02
Support signature verification for OCI image volumes


JasonPowr:
**Description**

Kubernetes v1.36 introduced a new volume type that allows mounting OCI images directly as volumes (spec.volumes[].image). Currently, the policy controller only verifies signatures on container images (containers, initContainers, ephemeralContainers) and does not inspect volume definitions.

This means an unsigned or untrusted OCI image can be mounted into a pod via an image volume without any policy enforcement.

The controller should validate image references in spec.volumes[].image.reference against ClusterImagePolicy rules, the same way it already does for container images.

PR #1976, opened 2026-06-05
feat: support validating oci volume mounts

JasonPowr:
#### Summary

- Extend the admission webhook to validate OCI image references in volumes[].image fields, enforcing the same signature and attestation policies already applied to container images
- Resolve volume image tags to digests in the mutating webhook, consistent with existing container image resolution behavior

Solves: https://github.com/sigstore/policy-controller/issues/1974

#### Release Note

feat: Add validation and resolution support for OCI image volume mounts

The admission webhook now validates image references in pod volume `image` sources, enforcing the same signature and attestation policies applied to container images. Volume image tags are also resolved to digests by the mutating webhook.

## OCI Image Volume Validation

The policy controller now validates OCI image references specified in Kubernetes volume mounts (`volumes[].image.reference`) in addition to container images.

When a pod spec includes a volume with an `image` source, the admission webhook will:

- **Validate** that the image reference is a digest (not a tag)
- **Enforce** any matching `ClusterImagePolicy` against the volume image
- **Resolve** image tags to digests via the mutating webhook (on create/update)

Non-image volumes (ConfigMap, Secret, EmptyDir, etc.) are unaffected.

### Example

A pod spec with an OCI image volume:

```yaml
spec:
  containers:
    - name: app
      image: registry.example.com/app@sha256:abc123...
  volumes:
    - name: data
      image:
        reference: registry.example.com/data@sha256:def456...

### DIFF
--- a/pkg/webhook/validator.go
+++ b/pkg/webhook/validator.go
@@ -347,9 +347,53 @@ func (v *Validator) validatePodSpec(ctx context.Context, namespace, kind, apiVer
 		wg.Wait()
 	}
 
+	checkVolumeImages := func(vols []corev1.Volume) {
+		results := make(chan containerCheckResult, len(vols))
+		wg := new(sync.WaitGroup)
+		count := 0
+		for i, vol := range vols {
+			if vol.Image == nil || vol.Image.Reference == "" {
+				continue
+			}
+			count++
+			i := i
+			ref := vol.Image.Reference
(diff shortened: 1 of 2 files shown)


sabre1041 (approved):
LGTM

Verified functionality of new integration

codysoyland (approved):

PR #1997, opened 2026-07-27
Refactor TestGetAuthorities into table-driven subtests

nellshamrell:
## Summary

Converts the monolithic `TestGetAuthorities` in `pkg/apis/config/image_policies_test.go` into named, table-driven subtests, resolving the long-standing in-code TODO:

```go
// TODO: Clean up this test to be table-driven with sub-tests, instead of one big test.
```

## Motivation

The original test ran ~15 independent scenarios sequentially against one shared config. A single `t.Fatalf` (e.g. via the `getAuthority` helper) aborted **all** remaining assertions, and failures reported generic messages with no indication of which scenario broke.

## Changes

- Each scenario now runs as its own `t.Run` subtest, so failures are isolated and clearly named (`--- FAIL: TestGetAuthorities/keyless_authority`, etc.).
- Homogeneous cases (single key-authority matches, resource matching) are grouped into small tables; heterogeneous cases (keyless, multiple-matches, attestations, source-oci, signature-pull-secrets) are individual named subtests.
- Panic-prone dereferences (nil `Keyless.InsecureIgnoreSCT`, indexing into `Identities`/`Sources`/`Attestations`/`SignaturePullSecrets`) are guarded so a malformed fixture yields a clean subtest failure instead of a panic.

## Behavior preservation

This is a **test-only, behavior-preserving** refactor: the same assertions run with the same expected values as before. No production code is touched.

## Testing

```
go test ./pkg/apis/config/
go vet ./pkg/apis/config/
gofmt -l pkg/apis/config/image_policies_test.go
```

All pass / clean.


nellshamrell:
Heads up: the failing CI on this PR is **not** related to these changes. It's a repo-wide breakage from an expired TUF `root.json` test fixture, tracked in #1999.

- The failures are all in `pkg/apis/policy/v1alpha1` (`TestTrustRootValidation`, `TestReconcile`) with `tuf refresh failed: expired metadata error: final root.json is expired`, plus the e2e jobs that depend on a healthy build.
- This PR only touches `pkg/apis/config/image_policies_test.go`; that package's unit tests pass, along with gofmt, goimports, lint, DCO, whitespace, license, and codegen checks.
- The same failures reproduce on `main` without this change.

This should go green once #1999 is resolved.

nellshamrell:
Hi @malancas! Anything I can do to help with the review on this one?

malancas (approved):
Lgtm, thanks for updating these tests

PR #2000, opened 2026-07-28
Regenerate expired TUF test fixtures

### LINKED ISSUE #1999: Unit + e2e tests failing on main: expired TUF root.json test fixture (validRepository)
## Description

Unit tests (and consequently the e2e jobs) are currently failing on `main` and on every open PR, because a hardcoded TUF repository test fixture has an embedded `root.json` whose expiry date has passed.

The embedded metadata expired on **2026-07-18T08:24:13Z**, so any test run after that date fails when it tries to construct a TUF client from the fixture.

## Failing tests

- `TestTrustRootValidation/Should_work_with_a_valid_repository` — `pkg/apis/policy/v1alpha1`
- `TestReconcile/With_repository` and `TestReconcile/With_repository_containing_trusted_root.json` — trustroot reconciler

All fail with:

```
failed to create TUF client: failed to load metadata:
tuf refresh failed: expired metadata error: final root.json is expired
```

## Root cause

`pkg/apis/policy/v1alpha1/trustroot_validation_test.go:32` defines `validRepository`, a base64-encoded, gzipped, tarred TUF repository committed as a static blob. Its embedded `repository/root.json` (and `1.root.json`) contain `"expires": "2026-07-18T08:24:13Z"`. Once that timestamp is in the past, `Compile`/TUF-client construction rejects the metadata as expired, so the fixture can no longer be loaded.

Decoding the blob confirms the expiry:

```
repository/1.root.json -> expires: 2026-07-18T08:24:13Z
repository/root.json    -> expires: 2026-07-18T08:24:13Z
```

## Impact

- `Run unit tests` fails on `main`.
- All `ClusterImagePolicy e2e tests` and CRD/e2e matrix jobs fail as a result.
- Every open PR is blocked by red CI regardless of its content.

## Suggested fix

Regenerate the `validRepository` fixture with a fresh (or far-future) expiry, and/or generate it at test time instead of committing a static blob with a baked-in expiry. The repo already has a TUF repository builder in `pkg/tuf/repo_test.go` (`GenKeyWithExpires`, `AddTargetsWithExpires`, `expires := time.Now().AddDate(0, 6, 0)`) that could serve as the basis for a non-expiring or dynamically generated fixture.

## Steps to reproduce

```
go test ./pkg/apis/policy/v1alpha1/ -run 'TestTrustRootValidation/Should_work_with_a_valid_repository'
```

nellshamrell:
## Summary

The TUF test fixtures across three packages shared a baked-in expiry of `2026-07-18T08:24:13Z`, which has now passed. Any test run after that date fails to construct a TUF client with `final root.json is expired`, reddening CI on `main` and every open PR.

Fixes #1999.

## Affected packages / tests

- `pkg/reconciler/trustroot` — `TestReconcile/With_repository` and the `trusted_root.json` variant (embedded `testdata/*`).
- `pkg/apis/policy/v1alpha1` — `TestTrustRootValidation` (inline base64 blobs).
- `pkg/tuf` — `TestDownloadTargetFromSerializedMirror` and `TestClientFromSerializedMirror` (inline base64 blobs, byte-identical to the v1alpha1 ones). *This third package was not listed in the original issue; found while preparing the fix.*

## Changes

- Regenerated `pkg/reconciler/trustroot/testdata/*` via `make generate-testdata`.
- Replaced the inline `validRepository` / `rootJSON` blobs in `pkg/tuf` and `pkg/apis/policy/v1alpha1` with the freshly generated `tufRepo.tar` and `root.json` (base64-encoded). All three fixtures now chain to the same fresh trust anchor.
- Refreshed the stale `next expiration` comments and replaced the manual kind-cluster generation notes with a pointer to `make generate-testdata`.

## Validation

- `go test ./pkg/tuf/... ./pkg/apis/policy/v1alpha1/... ./pkg/reconciler/trustroot/...` — pass
- `go test ./...` — pass
- `gofmt` / `go vet` — clean

## Note

The scaffolding repo builder hardcodes a 6-month expiry, so these fixtures will need periodic regeneration (next: `2027-01-28`). A longer-lived or clock-injectable fixture would avoid the recurrence; happy to follow up separately if maintainers prefer that direction.


nellshamrell:
Filed #2001 as a follow-up to address the recurring 6-month expiry so this stops needing periodic regeneration.

nellshamrell:
## Note on the failing e2e checks

The failing jobs on this PR are **all e2e**, and they fail at the `Install policy-controller` build step (before any test runs), in `make ko-policy-controller`:

```
tarring kodata: kodata symlink ".../cmd/webhook/kodata/HEAD"
resolves to ".../.git/HEAD" which is outside the kodata root
make: *** [Makefile:141: ko-policy-controller] Error 1
```

This is a **pre-existing, repo-wide infrastructure failure**, not caused by this change:

- This PR only modifies `_test.go` files and `testdata/` binaries — nothing in the build path.
- The **identical** error occurs on unrelated PRs (e.g. #1997 and the open dependabot PRs).
- The `cmd/webhook/kodata/{HEAD,refs}` symlinks point into `.git/` and date from 2022 (unchanged). A newer `ko` now rejects symlinks that resolve outside the kodata root, which breaks `ko resolve` for the whole repo.

The unit-test layer this PR targets is green (`Run unit tests`, gofmt, goimports, codegen, lint, DCO all pass). The e2e breakage needs a separate infra fix (the `ko`/kodata symlinks); I'll open a separate PR for that and link it here.

nellshamrell:
Follow-up: I've opened #2002 to fix the repo-wide e2e/build CI breakage referenced above. The failures on this PR (`tarring kodata: ... resolves to .git/HEAD which is outside the kodata root`) come from ko v0.19.0's new kodata-symlink check (ko-build/ko#1619) combined with `setup-ko` defaulting to `latest-release`; #2002 pins ko to v0.18.1 until the upstream relaxation (ko-build/ko#1699) is released. Once #2002 merges, re-running e2e here should get past `ko resolve`.

nellshamrell:
Closing in favor of pulling these changes into #2002

PR #2002, opened 2026-07-28
Pin ko to v0.18.1 in CI to fix kodata symlink failures

nellshamrell:
## Summary

Every e2e and build workflow currently fails at the **Install policy-controller** step (`make ko-policy-controller` → `ko resolve`) with:

```
tarring kodata: kodata symlink ".../cmd/webhook/kodata/HEAD" resolves to
".../.git/HEAD" which is outside the kodata root ".../cmd/webhook/kodata"
```

This affects **all branches and PRs** (dependabot, and open PRs like #1997 and #2000), not any individual change.

## Root cause

The workflows install ko via `ko-build/setup-ko` **without a `version:` input**, so they use setup-ko's default `latest-release`. ko **v0.19.0** (ko-build/ko#1619) added a strict check that rejects kodata symlinks whose targets resolve **outside** the kodata directory. `cmd/webhook/kodata/` contains such symlinks (present since 2022):

- `HEAD` → `../../../.git/HEAD`
- `refs` → `../../../.git/refs`
- `LICENSE` → `../../../LICENSE`
- `VENDOR-LICENSE` → `../../../third_party/VENDOR-LICENSE`

So when CI silently moved to ko ≥ v0.19.0, `ko resolve` began failing repo-wide.

The upstream relaxation ko-build/ko#1699 ("allow kodata symlinks within the source tree") is **merged but not yet released** (latest release is v0.19.1; the fix landed after it).

## Fix

Pin `setup-ko` to **ko v0.18.1** (the last release before the strict check) across all workflows that install ko, with an explanatory comment. This is an **explicit, temporary** workaround: v0.19.0's check is a legitimate hardening (it blocks kodata symlinks from packing arbitrary host files into images), so the pin should be **removed once a ko release includes #1699**.

## Validation

Reproduced locally with the real ko binaries against `./cmd/webhook`:

- **ko v0.19.1** → fails at kodata tarring with the exact CI error.
- **ko v0.18.1** → packs kodata successfully (only fails later at publish, because no local kind cluster — unrelated to kodata).
- Removing **only** the `.git` symlinks is **insufficient**: ko v0.19.x then rejects `LICENSE` next. Pinning is the minimal correct fix.

## Notes

- This unblocks e2e for all currently-open PRs. e2e may still surface unrelated failures further along, but it will get past `ko resolve`.
- Stale `# will use the latest release available for ko` comments are removed, since ko is now pinned.
- A follow-up to remove the pin once ko#1699 ships would be appropriate.


nellshamrell:
Update: with the ko pin in place, the approved run showed the e2e suite passing again — the only remaining failures were `Run unit tests` and the TrustRoot "Bring Your Own Keys" e2e, both caused by the expired TUF test fixtures fixed in #2000. I've cherry-picked #2000's `Regenerate expired TUF test fixtures` commit onto this branch so #2002 is green on its own. Verified locally: `go test ./pkg/tuf/... ./pkg/reconciler/trustroot/... ./pkg/apis/policy/v1alpha1/...` all pass. (This does overlap with #2000; whichever merges second will simply no-op those fixture changes.)

malancas (approved):

PR #2004, opened 2026-08-05
Generate TUF test fixtures at test time instead of committing them

### LINKED ISSUE #2001: TUF test fixtures recur every 6 months due to hardcoded scaffolding expiry
## Background

PR #2000 (Fixes #1999) regenerated the expired TUF test fixtures across three packages:

- `pkg/reconciler/trustroot` (embedded `testdata/*`)
- `pkg/apis/policy/v1alpha1` (inline `validRepository` / `rootJSON` blobs)
- `pkg/tuf` (inline `validRepository` / `rootJSON` blobs)

That unblocks CI, but it is a recurring papercut: the fixtures now expire on **2027-01-28**, and CI will go red again then, on `main` and every open PR, with `final root.json is expired`.

## Root cause

The fixtures are generated by `hack/gentestdata/gentestdata.go` (`make generate-testdata`), which calls `scaffolding/pkg/repo.CreateRepoWithOptions`. Scaffolding hardcodes:

```go
// scaffolding/pkg/repo/repo.go
expires := time.Now().AddDate(0, 6, 0) // 6 months
```

So every regeneration is only valid for 6 months, and there is no knob to extend it. The same expiry is baked into the committed fixtures, so the whole cycle repeats.

## Proposed options (in rough order of preference)

1. **Generate fixtures at test time** for the packages that can support it (e.g. `pkg/tuf` and `pkg/apis/policy/v1alpha1`), so `now + 6mo` is always in the future and the fixtures never expire in CI. `pkg/tuf/repo_test.go` already contains a `createRepo` helper using go-tuf's `*WithExpires` APIs that could be the basis. Caveat: scaffolding's `CreateRepoWithOptions` writes to a fixed `/tmp/tuf` path, so a runtime approach should use `t.TempDir()` / an in-memory store to avoid cross-package collisions.
2. **Inject a controllable clock / expiry** into the TUF client validation used by tests, so expiry can be pinned deterministically.
3. **Upstream a configurable expiry** in `scaffolding/pkg/repo` (e.g. a `CreateRepoOptions.Expires` field), then set a long-lived expiry in `gentestdata.go`. This also benefits other scaffolding consumers.
4. **Interim mitigation:** a scheduled CI job (or a documented calendar reminder) that runs `make generate-testdata` + re-encodes the inline blobs and opens a refresh PR before each expiry.

The golden-comparison fixtures in `pkg/reconciler/trustroot` are harder to move to pure runtime generation (they're compared against committed `marshalledEntry.json` etc.), so options 2–4 may fit those better than option 1.

## Acceptance

- TUF-fixture-dependent tests no longer fail purely due to the passage of time, or the recurrence is reduced to an automated, low-effort refresh.

Follow-up to #1999 / #2000.

nellshamrell:
#### Summary

TUF test fixtures in this repo were committed as base64 blobs / tarballs whose
embedded metadata expires 6 months after generation, because
`scaffolding/pkg/repo.CreateRepoWithMetadata` hardcodes `time.Now().AddDate(0, 6, 0)`
with no way to configure it. These fixtures were already regenerated once
(shipped via #2002) and, left alone, would expire again on 2027-01-28 and
redden CI purely from the passage of time.

This replaces the committed fixtures with a shared test-only helper,
`internal/tuftest`, that signs a fresh TUF repository at test time with an
expiry relative to `now`, so it is never stale. It's used in `pkg/tuf`,
`pkg/apis/policy/v1alpha1`, and `pkg/reconciler/trustroot` (the three packages
affected). Six now-dead fixture files are removed from
`pkg/reconciler/trustroot/testdata`, and `hack/gentestdata` no longer generates
TUF repos (it still generates the keys/certs/golden JSON that remain
committed, since the reconciler's golden output is derived from those, not
from the TUF metadata).

Also adds a guardrail test (`internal/tuftest/guard_test.go`) that scans the
tree for TUF metadata in any of the shapes these fixtures have historically
taken - JSON metadata files, tarred repositories, and base64 blobs inlined in
Go source - so an expiring fixture can't quietly reappear. It has its own
tests covering all three shapes.

How to test: `go test ./...`, `go vet ./...`, `golangci-lint run` all pass. I
additionally stress-tested the previously-flaky expiry test 15 times in a row
(clean) and mutation-tested the guardrail with the real historical fixture
blobs from git history to confirm it actually detects them.

Fixes #2001

#### Release Note

NONE

#### Documentation

No user-facing behavior changes; this only affects test fixtures.


nellshamrell:
Feel free to close if this is not useful

Hayden-IO:
@jku fyi, thought you might find this interesting

nellshamrell:
Hello @jku! Anything I can do to help with the review of this one? (I'm also in the Sigstore Slack)

malancas (approved):

